PR #1, opened 2017-09-11
feat: add geckopy python package

hredestig:
hi @BenjaSanchez, opening this still very much work-in-progress pull request to let you review general strategy and things like the proposed package name

Scope of the package is to be able to get an adjusted cobra model given proteomics measurements.

CI here https://travis-ci.org/DD-DeCaF/GECKO


BenjaSanchez:
Dear @hredestig,
I have some questions before I merge your branch to the master one:
1. Could the new package instead of _cobra_gecko_ be named _limit_proteins_ and be contained inside the _Python_Module_? This to be consistent with the rest of the package.
2. I noticed that there is a readme file and a license. Do those have to stay? The main already has both of them.
3. You have stored the models in a separate folder called _data_files_ but ideally they should be stored just once to avoid confusion. Is there a need to have those files there and not just use the ones in _Models_? The file _enzymes.txt_ could also be transferred there if you want (although it is not evident to me why it has 6705 rows if there are only 764 enzymes).
4. Finally and most importantly, would you mind if I make these changes myself after merging or do you prefer to first make them yourself in your branch and then send a new pull request?

Let me know then how to proceed

BenjaSanchez:
no problem, then I'll wait for you to finish up. The name `cobra_gecko` implies that the rest of the GECKO toolbox is not meant for COBRA, hence it will be confusing for users. If you want it can be called `GECKOpy` or similar, let me know if you have a better name suggestion :)

hredestig:
I fixed your original points and renamed the package. Superficially it works in my hands now but haven't yet written tests for checking that simulations are correct, and, in particular that biomass and pool constraints need review and some tests. But you can go ahead an merge now if you like and we take that in further PRs.

hredestig:
Note I made a bit re-formatting of README.md, hope you don't mind, please adjust as you see fit. Unfortunately however, PyPI doesn't support markdown on the package pages so as I'm copying like this  it will  look messed up there. The easiest option would be to change to '.rst' instead, I can do that if you like.

Also, any particular reason you are using your own license instead of some common public license, e.g. Apache?

hredestig:
And, what would be a sensible default for `P_total` in case this wasn't measured?

hredestig:
I tried to make tests to verify that I could get the same constants etc as you report in the SI / fetched from the model you provided, alas this is not the case. [Quite substantial differences](https://github.com/DD-DeCaF/GECKO/blob/feat/python-package/geckopy/tests/test_gecko.py#L22) but can't really see why. Perhaps you can now review this and see if you can find any mistakes..

Midnighter:
@BenjaSanchez as Henning has moved on to new challenges, I'm taking stewardship of developing the Python end of the GECKO package. Please let me know if there is work required on my end or whether you just need some more time to review this.

BenjaSanchez:
Dear @hredestig and @Midnighter,
Sorry for the delay in the reply. I have gone through the latest pull and it looks good to merge. A couple of additional comments:
1. Regarding the README file, it would be nice to not loose compatibility with the Github Page I created. So I will adapt the file to .rst and let you know if I need some assistance with that :)
2. Regarding the license, I used a standard MIT license, and I just added the condition for citing the software. Do let me know if you think something is amiss, any advantage of using Apache?
3. Regarding the differences that you see with the original model, note that the current model has changed considerably from the one in the publication (additional manual curation and improved matching algorithm). If you want to properly compare you need to use the model from the [1st release](https://github.com/SysBioChalmers/GECKO/releases/tag/v1.0) of the package. If there are still differences let me know and I can look further into it.

Midnighter:
> So I will adapt the file to .rst and let you know if I need some assistance with that

You can start from `pandoc README.md -o README.rst` it works pretty well.

> any advantage of using Apache?

MIT and Apache are both very liberal licenses. I don't think either of our organizations is set up to enforce licenses anyway but the Apache license has the "scare" factor that if you break the license terms you are bared from using any Apache licensed software ever again. So that's a pretty big deterrent.

BenjaSanchez:
thanks! will try it out and let you know if I have any trouble

PR #7, opened 2017-12-02
fix: include data in Python package

Midnighter:
* prevent Travis from removing the models before distribution
* add Midnighter to contributors
* include the complete `data_files` directory in the Python package
  distribution
* remove unused AppVeyor config


Midnighter:
Travis recently updated their Mac images with a new XCode version which seems to mess up many configurations. :disappointed:

PR #38, opened 2018-08-01
Feat/RAVENdependency

BenjaSanchez:
Added [RAVEN](https://github.com/SysBioChalmers/RAVEN) as a dependency for the Matlab scripts, to avoid redundancy of functions `convertToIrreversibleModel` and `standardizeGrRules`. No model changes, except for the number of decimals in the `.txt` file due to some changes in COBRA, most likely from https://github.com/opencobra/cobratoolbox/pull/1294 (however, no changes to the `.xml` file)


BenjaSanchez:
@IVANDOMENZAIN let me know if you see any changes in the `.txt` files after running the pipeline in MAC

IVANDOMENZAIN:
Hej Ben,

I ran the whole pipeline and everything's alright on the .xml files, there are some differences on the stoichiometric coefficients (on the 5th or 6th significant digits) for the .txt files, but overall those were the only changes.

BenjaSanchez:
@IVANDOMENZAIN we will have to create a new branch to address those differences then, we don't want changes when moving from one OS to another. This also needs to be solved in https://github.com/SysBioChalmers/yeast-GEM/pull/143.

IVANDOMENZAIN:
No changes in the .txt files after the RAVEN and COBRA updates, everything's in its right place!

PR #43, opened 2018-09-11
Fix add enzymes to rxn

IVANDOMENZAIN:
It was found that the GECKO pipeline matches each gene in the model with its first associated protein in a given database (swissprot or KEGG), however when multiple IDs are available for the queried protein the current pipeline just ignores the additional entries. In this branch the issue has been addressed in the next way:

- The `getEnzymeCodes` function can be called with optional arguments for: I) ignoring multiplicity cases; II) Display all the genes with multiple proteins to the user, and III) Adding additional matches 
as isoenzymes for the given reaction.

- `findInDB`saves the exact gene that was matched to each protein and this information is included now in the `model_data` data structure which avoids possible ambiguities in further steps in the pipeline.

These solutions are particularly helpful when dealing with human-based GEMs in which splice variant proteins might be present.


IVANDOMENZAIN:
Looks fine! everything's working properly

Issue #15, opened 2018-03-15
drop .mat storage


BenjaSanchez:
all .mat files in the repo should be changed to flat files, to avoid conflicts

PR #52, opened 2018-10-25
Fix/remove binaries

BenjaSanchez:
Adressing issue #15, most binaries of the model have been removed:

* `ecYeastGEM.mat` & `ecYeastGEM_batch.mat` are not stored anymore in GECKO (but are available in the [ecModels container](https://github.com/SysBioChalmers/ecModels/tree/master/ecYeastGEM/model)).
* `enzData.mat` is also not stored anymore (out of use). The `kcats` & `model_data` outputs of `enhanceGEM.m` have been replaced by `ecModel_batch` (which will be needed later by ecModels).
* Moved `kcatModifications.txt` & `topUsedEnzymes.txt` out of the "data" subfolder, for simplification.

Only binaries remaining are `protDatabase.mat` and `PhylDist.mat`, (the former will not change anymore and the latter will seldom change)


edkerk:
As you mention `ecModel_batch` in your commit message, this name seems confusing, is this model to simulating batch experiments while the other model is for chemostats? Wouldn't `_onePool` or something like that be more descriptive, referring to the fact that no proteomics data is integrated?

BenjaSanchez:
@edkerk `ecModel_batch` does simulate batch conditions, i.e. unlimited growth on a single carbon source, while `ecModel` simulates no conditions at all. I agree that it might be slightly confusing, but atm it is kept for BC reasons. In any case, that can be discussed further in a separate issue, I think it would be fine to either change name or document better :)

PR #53, opened 2018-10-29
Fix/scaleBioMass

BenjaSanchez:
* Some fixes in the biomass composition + GAM calculations.
* As a result of the above, increased tolerance in kcat flexibilization from 5% to 10%.
* Added `sumProtein.m` for easier use when creating new ecModels.
* Documented better which scripts/data should be changed and which are optional when adapting geckomat to produce a new ecModel.


BenjaSanchez:
@IVANDOMENZAIN I do not think using the highest kcat value for all the isoenzymes in a reaction will force the model to split the flux, as when using pFBA, the simulation would then choose the isozyme with the lowest MW (as the overall activity kcat/MW will be highest). Isozymes have been a limitation of GECKO from the start, and unless we provide proteomics data (where different isozymes levels might create different limitations) I don't see any way of addressing this with the current formalism, but I'm `optimistic` that future enhancements of GECKO (for instance with Km data) would make different isozymes chosen under different conditions.

Issue #51, opened 2018-10-23
improve usability


edkerk:
It is not that straight-forward to generate models, integrate proteomics data and perform simulations with gecko. I have recently run into the following hurdles:

- Lack of documentation. The [appendix](http://msb.embopress.org/content/msb/13/8/935/DC1/embed/inline-supplementary-material-1.pdf?download=true) to the gecko paper is helpful, but doesn't cover all functions, while the [readthedocs](https://geckotoolbox.readthedocs.io/en/latest/) provides only minimal information.
- Hard-coded model-specific parameters, such as reaction names and IDs, organism name, medium composition, growth simulations, fitting of GAM etc.
- Hard-coded locations of files and scripts.
- Nested functions, such as `simulateChemostat`
- Related to the points above, `enhanceGEM` works well to generate ecYeast, but might be largely unsuitable for other models. 

Besides the documentation, probably a lot can be improved by defining global variables. This includes things like reaction IDs for substrate uptake, location of data files _etc_. This would also simplify the nested functions, as parameters can be accessed by each function. It's imperative that model-specific functions exist, such as `simulateChemostat` and `manualModifications`, but this way it would be easier to adjust those for wider applicability.

BenjaSanchez:
@edkerk thanks for pointing out these problems. Below some answers:

> Lack of documentation

We will be working soon on providing more documentation, let us know which parts are the ones that could use more work.

> Hard-coded model-specific parameters

A new PR will be made this week including a better way of handling `Pbase` and `Ptot`. We will also clarify in the [usage](https://geckotoolbox.readthedocs.io/en/latest/readme.html#usage) section of the documentation which are the functions that should be replaced by the user and which can be ignored if desired (such as `fitGAM.m`, which can remain unchanged if `scaleBioMass.m` is modified to not call it).

> Hard-coded locations

All location paths of files/scripts are relative to from where they are called, so you should not have any problems with this, as long as you don't change the directory in your modifications. Let us know if this is not the case for any specific file/script.

> Nested functions

`simulateChemostat` is only used for fitting GAM so I don't think it's a problem that it's nested, as `fitGAM.m` can be modified by the user anyways.

> Other models

Note that we will be migrating all models to `ecModels` (currently to a large extent WIP), which will make `enhanceGEM.m` work for any of the models that we will host there.

> Global variables

[...]

PR #60, opened 2018-12-05
refactor: getConstrainedModel uses constrainPool directly

edkerk:
- refactor:
  - instead of parsing through `constrainEnzymes`, which is confusing as no enzymes are individually constrained, directly call `constrainPool`
  - `constrainPool` has alternative option where all enzymes are assumed non-measured 

To me it has been confusing that `constrainEnzymes` is called when only the pool is constrained. Perhaps there are good reasons for this, in that case please discard this PR.


BenjaSanchez:
@edkerk `constrainEnzymes.m` is also used when constraining the model with proteomic data, it has inputs `pIDs` & `data` for this. That being said:
1. We should improve the documentation of both `constrainEnzymes.m` and `constrainPool.m` to explain this (connected to https://github.com/SysBioChalmers/GECKO/issues/51).
2. We should also add an example in the tool showing an integration with proteomics data, for instance in `/geckomat/utilities`, or possibly in a `/tutorials` folder.

BenjaSanchez:
@edkerk the reason in `getConstrainedModel.m` we call `constrainEnzymes.m` and not `constrainPool.m` is because in the former we also perform some other changes, e.g. `scaleBioMass` and `fitGAM`.

edkerk:
Just to clarify, I don't advocate for discarding `constrainEnzymes`, but just to directly call `constrainPool` by `getConstrainedModel` when it is generating the `ecModel_batch` model. I don't see why this has to run _through_ `constrainEnzymes`

BenjaSanchez:
@edkerk the reason in `getConstrainedModel.m` we call `constrainEnzymes.m` and not `constrainPool.m` is because in the former we also perform some other changes, e.g. `scaleBioMass` and `fitGAM`.

PR #61, opened 2018-12-05
refactor/feat: topUsedEnzymes, truncateValues and modifyKcats

edkerk:
- doc:
  - explanation of input parameters
- refactor:
  - `topUsedEnzymes`only generate table from cell array near end of function, for speed improvement and getting rid of warnings when new data was added to the table
- feat:
  - `topUsedEnzymes` optional parameters: `writeFile` and `n` for number of top used enzymes, and now supporting `T` as command line output
  - `topUsedEnzymes` output table now contains additional column specifying the condition, this changes the output from before this commit
  - `truncateValues` directly uses column numbers: instead of `nCols` referring to last number of columns, `cols` now contains directly which columns should have their value truncated.
- fix:
  - `modifyKcats` supports new `truncateValues` behaviour


IVANDOMENZAIN:
Ready to merge now!

PR #62, opened 2018-12-07
feat: various minor fixes and feats

edkerk:
- feat
  - `updateDatabase` keggId is a parameter to be provided, instead of hard-coding _S. cerevisiae_
- fix
  - `sigmaFitter` if no solution can be found, give vector of zero instead of error
  - `measureAbundance` ignore NaN that could be part of relative proteomics dataset
  - `manualModifications` correctly adjusts aconitase kcat
- style
  - reduce number and ensure consistent style of progress messages from building `updateDatabase`, `manualModifications`, `measureAbundance`, `matchKcats`, `getEnzymeCodes`


BenjaSanchez:
@edkerk which editor do you use when working on matlab scripts? For 6/7 of the modified files, line endings changed with`^M` at the end, making diffing between files [very hard](https://github.com/SysBioChalmers/GECKO/pull/62/files). This is because of different formats between linux and windows line endings, and can be quickly fixed using the CLI tool dos2unix like this: `dos2unix fileName.m`. See more [here](https://stackoverflow.com/questions/64749/m-character-at-end-of-lines). Could you do that in this PR so I can properly review it?

edkerk:
Seems like Notepad++ introduced those characters. I've recently switched to Visual Studio Code, which immediately shows whether Unix or Windows EOL are used. I noticed some more functions having those characters (some of those introduced by me, I'm afraid), I will fix that in a separate PR to keep this clean.

(Note that getEnzymeCodes already had the ^M codes, the real difference that I'm introducing here is in line 152-154)

PR #69, opened 2019-02-08
fix/comparative fva

IVANDOMENZAIN:
It was found that on the function `MAX_min_Optimizer` the minimization step was just done for the forward reaction (irrevModels), the bug has been fixed and now variability ranges are calculated properly for both directions (forward and backward) of a given reaction in an irreversible model.


edkerk:
@IVANDOMENZAIN Regarding excluding blocked reactions, would it maybe be faster to replace (~line 35):

```matlab
rxnsIndxs = find(model.c~=1);
```
with
```matlab
I=haveFlux(model);
rxnsIndxs = find(model.c~=1 & I);
```
?

IVANDOMENZAIN:
@edkerk You were right, I already implemented and tested your idea and it worked well. @BenjaSanchez This is ready to be merged!

PR #82, opened 2019-09-10
Feat/Incorporate_proteomics

IVANDOMENZAIN:
A family of functions for automated generation of ecModels with proteomics constraints has been added. For a given dataset with semi-absolute protein abundances  [mmol/ gDw] for **m** replicates for **n** conditions, median values are obtained for each protein in each condition and then added as individual constraints to the enzyme usage pseudoreactions in the **ecModel** structure, allowing the construction of a protein-constrained condition-specific **ecModel**. 

The constraints incorporation algorithm consists of the following steps:
1. Load fermentation data. This should be provided in a tab separated **.txt** file named **GECKO/Databases/fermentationData.txt**. This file contains Dilution rate, average total protein content per condition, glucose, CO2 and Oxygen exchange fluxes  in [mmol/gDw h]. Each row contains data entries for a given experimental condition. As an option, the user can provide byproducts exchange fluxes in additional columns using the same units and the model's metNames for the respective byproducts as column names. An example file can be seen at:  2dbf420. The **proteomics dataset** is also loaded in this first step, it should be provided by the user by storing it with the name **GECKO/Databases/abs_proteomics.txt** as a tab delimited file. An example of the expected format can be seen at: 
b046047.

2. For each experimental condition, the proteomics dataset is filtered, excluding proteins that are not present in at least 2/3 of the total number of condition replicates and also noisy measurements (relative standard deviation higher than 1).

3. Proteomics abundances are corrected for the oxidative phosphorylation complexes, trying to avoid overconstraining of potentially erroneously measured subunits that might limit the whole pathway. This correction is limited to just this pathway as it is desirable to modify the original dataset the least possible and also, abundance changes in OxPhos compounds are key to meet the phenotype energy requirements.

4. Set medium constraints. Allows the uptake of whatever compound present in the experimental culture medium.

5. If the condition-dependent total protein content differs in more than 5% of the original **ecModel** total protein content, then biomass composition is rescaled and GAM is fitted according to the chemostat data provided in [GECKO/databases/chemostatData.tsv](https://github.com/SysBioChalmers/GECKO/blob/master/databases/chemostatData.tsv).

6. Byproducts production rates are set as upper bounds for their respective exchange reactions in the model.

7. Using an **ecModel_batch** with the same constraints setup, minimal enzyme requirements for the proteins present in the filtered dataset are retrieved from a FBA solution vector. Enzyme abundances that are lower than the minimum requirements calculated by the FBA solution are corrected in the dataset.

8. A proteomics constrained **ecModel** is obtained by the function **`constrainEnzymes.m`**. A flexibilized carbon source uptake rate is set as upper bound (1.05*experimental value) and then **`flexibilizeProteins.m`** flexibilizes the top-limiting abundances until the model is able to grow at the provided experimental growth rate. After this,  an optimal enzyme usage profile compatible with the provided constraints is obtained and optimal levels are set as upper bounds for the flexibilized protein usages.

9. The total flexibilized mass of protein is drawn from the remaining protein pool (upper bound for `protein_pool_exchange` rxn) for consistency with mass conservation.

10. Non-growth associated ATP maintenance is fitted according to the condition specific experimental data.

11. Chemostat conditions are set by first fixing growth rate to the experimental value, minimizing the carbon source uptake, fixing its optimal value and then setting the total unmeasured enzymes usage as a new objective to minimize. 

12. Each condition-specific model is saved in **GECKO/models/prot_constrained**


IVANDOMENZAIN:
@BenjaSanchez I think that this PR is ready now. I'm sorry for making it this long, several ideas needed to be tested in order to achieve a general series of steps.

IVANDOMENZAIN:
> * flex_mass

Yes! it should be zero.

IVANDOMENZAIN:
> Great! I have no more comment. But if the usage of generating a proteome-constrained model is written somewhere for example readme file, so when I am not familiar with GECKO repo structure, then that could save some time scanning all the functions. If I am wrong and there is actually a guideline for that, ignore me.

I agree with you @feiranl, a continuously updated README that describes all of the utilities scripts can be added, it might be also a good idea to add this to the `READtheDocs` website. What do you think @BenjaSanchez?

BenjaSanchez:
@IVANDOMENZAIN @feiranl I agree that more documentation for these utilities would be great. To not make even lengthier this PR though (over 50 commits already), how about we address that in a separate PR? Issue #88 will remain open until then.

BenjaSanchez:
@IVANDOMENZAIN I have added a few fixes to keep compliance with the original `enhanceGEM` pipeline. Just to confirm, did you do the proteomics integration example with the ecModels of yeast 8.3.2 available at [ecModels](https://github.com/SysBioChalmers/ecModels/tree/19049509fc374d92da157e2eaf9917f90ef2c230/ecYeastGEM/model)? When running it in my setup, I get slightly different values in most output files, but this might be because of different machine/solver setups.

IVANDOMENZAIN:
> @IVANDOMENZAIN I have added a few fixes to keep compliance with the original `enhanceGEM` pipeline. Just to confirm, did you do the proteomics integration example with the ecModels of yeast 8.3.2 available at [ecModels](https://github.com/SysBioChalmers/ecModels/tree/19049509fc374d92da157e2eaf9917f90ef2c230/ecYeastGEM/model)? When running it in my setup, I get slightly different values in most output files, but this might be because of different machine/solver setups.

@BenjaSanchez Thanks for pointing this out! I actually haven't updated the results ouput files after the last additions, let me do it and commit them using the same model version as you and then we can see if it is a solver related issue or not.

IVANDOMENZAIN:
@BenjaSanchez, I just uploaded the updated results with the ecModel that is available in ecModels, however it is ecYeastGEM v8.3.4

PR #95, opened 2020-02-03
Fix/comparative_FVA

IVANDOMENZAIN:
Additional changes and fixes have been introduced to the comparative FVA algorithm:

1. Simplification of `model` and `ecModel` constraining procedure.
2. For reversible reactions, backward reactions are completely blocked when the correspondent forward reaction is being optimized and vice versa.
3. For the batch option,`Model` and `ecModel` can be constrained with the same suboptimal growth rate or with their correspondent optimal values.

These changes were introduced to perform flux variability analysis on 11 human cancer cell-lines models with enzyme constraints. Models, scripts and results available at:
https://github.com/SysBioChalmers/EnzymeConstrained_humanModels

### DIFF
--- a/geckomat/utilities/FVA/comparativeFVA.m
+++ b/geckomat/utilities/FVA/comparativeFVA.m
@@ -128,7 +128,7 @@ else
 end
 
 model.lb(objIndx) = factor*priorValue;
-model.ub(objIndx) = priorValue;
+model.ub(objIndx) = 1.0001*priorValue;
 
 sol = solveLP(model);
 if ~isempty(sol.f)
--- a/geckomat/utilities/FVA/comparativeFVA.m
+++ b/geckomat/utilities/FVA/comparativeFVA.m
@@ -2,67 +2,72 @@ function [FVA_Dists,indexes,blocked,stats] = comparativeFVA(model,ecModel,c_sour
 % comparativeFVA
 %  
 % This function goes through each of the rxns in a metabolic model and
-% gets its flux variability range, then the rxn is mapped into the EC
-% version of the model to perform the correspondent variability analysis and
+% gets its flux variability range, then the rxn is mapped into an EC
+% version of it to perform the correspondent variability analysis and
 % finally compares and plots the cumulative flux variability distributions. 
 %
 %   model       MATLAB GEM structure (reversible model), constrained with 
-%               the desired culture medium constrains
+%               the desired culture medium constraints and biomass
+%               pseudorxn as an objective function.
 %   ecModel     MATLAB ecGEM structure, constrained with the desired culture
-%               medium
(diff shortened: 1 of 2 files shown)


BenjaSanchez:
@IVANDOMENZAIN planning on reviewing this PR and #82 in the coming 2 weeks, sorry for the massive delay. Are they both ready to go?

IVANDOMENZAIN:
@BenjaSanchez, no worries there's no rush. This one is ready and it was already use in a study that has already been submitted. You better start with this one, I'm working with proteomics datasets now, so I would probably add some changes to #82 next week.

geckomat/utilities/FVA/comparativeFVA.m
```diff
@@ -109,52 +123,27 @@
...
-    irrevModel.lb(objIndx) = factor*priorValue;
-    irrevModel.ub(objIndx) = priorValue;
-else
-    sol = solveLP(irrevModel);
-    irrevModel.lb(objIndx) = factor*sol.x(objIndx);
-    irrevModel.ub(objIndx) = sol.x(objIndx);
-end
-sol = solveLP(irrevModel);
+model.lb(objIndx) = factor*priorValue;
+model.ub(objIndx) = 1.0001*priorValue;
```
BenjaSanchez, line 141: shouldn't this line only apply when `fixed=True`? As in other cases it could just be:
```matlab
model.ub(objIndx) = priorValue;
```
Or is it needed in all cases for simulation stability?

IVANDOMENZAIN:
> changes look great! Left a small question in one unclear part. My only other comment is that the [link](https://github.com/SysBioChalmers/EnzymeConstrained_humanModels) you are pointing to in the description seems to be broken, or is it a private repo?

You're right @BenjaSanchez this is a private repo of SysBio.

geckomat/utilities/FVA/comparativeFVA.m
IVANDOMENZAIN: When `fixed = True` the lower bound is also set, otherwise just an UB is set

BenjaSanchez: What I meant here is that, as I see it, when `fixed = True` it should be:
```matlab
model.lb(objIndx) = 0.9999*priorValue;
model.ub(objIndx) = 1.0001*priorValue;
```
whereas when `fixed=False` it should be:
```matlab
model.lb(objIndx) = 0;
model.ub(objIndx) = priorValue;
```
However, currently the `1.0001*priorValue` applies regardless of the value that `fixed` takes. Is that decision intentional or an oversight?

IVANDOMENZAIN: It was intentional @BenjaSanchez :)

PR #76, opened 2019-07-13
Refactor/avoid hard coding

IVANDOMENZAIN:
Avoid hard-coding of functions involved in the `getConstrainedModel` step of the pipeline. Model specific parameters are assigned by a global script (**`getModelParameters`**) that can be called every time that these parameters are needed in the whole GECKO workflow. Such parameters being the following:

- Sigma factor (Average enzyme saturation factor)
- Ptot (Total protein content in the cell [g protein/gDw])
- gR_exp (Minimum growth rate the model should grow at [1/h])
- org_name  (organism scientific name)
- keggID (organism KEGG ID)
- c_source (The name of the exchange reaction that supplies the model with carbon)
- bioRxn (Rxn Id for biomass pseudoreaction)
- enzyme_comp (Compartment name in which the added enzymes should be located)

This will allow a further generalization of the method, minimizing the amount of organism specific scripts that a user should create in order to generate a new ecModel (for a different organism) with the [ecModels generator pipeline](https://github.com/SysBioChalmers/ecModels).

Issue #55, opened 2018-11-07
hard-coded data file & script locations


edkerk:
In `measureAbundance`, the location of the protemics data is hard-coded. This is inconvenient if one would for instance have proteomics data for many conditions and want to make models for each of these conditions: the user would have to replace the `databases/prot_abundance.txt` file. Probably better to have `genes` and `abundance` as parameters to the function.

BenjaSanchez:
@edkerk note that `measureAbundance` is not meant for constraining the model with proteomics, but to read Pax-DB data and estimate the fraction [g/g] of a group of enzymes respect to the total, which is needed to (later) set a constrain on the protein pool for enzymes which don't have proteomic measurement. The function you would want to use for the purpose you describe is `constrainEnzymes`, which accepts inputs of `pIDs` and `data` as you suggest.

That being said, I agree with the problem of hard-coded locations (e.g. of `prot_abundance.txt`), and we will change this at the toolbox level. For this the idea would be to add `/geckomat*` and `/databases*` paths as a requirement for using GECKO, that way we can avoid any relative path and use those functions/data from any other folder. Let us know if you have any thoughts on this proposal.

edkerk:
Thanks for the explanation, this could go directly in the documentation! :) What if no Pax-DB data is available for my organism of interest?

Requiring defining the location of the `/geckomat*` and `/databases*` paths as parameters sounds like a good solution.

BenjaSanchez:
> What if no Pax-DB data is available for my organism of interest?

I guess it would have to be replaced with some proxy, e.g. the fraction of enzymes from the total, although that would underestimate metabolic enzymes... @IVANDOMENZAIN any ideas here?

IVANDOMENZAIN:
I have used relative proteomics datasets (when available) as a substitution for the `prot_abundance.txt` file, the `f` values that I have obtained for different organisms with this approach range from 0.3 to 0.48. As the `f` value is used for constraining the protein pool, I think that using a high value such as 0.48 or 0.5 also makes some sense because the protein pool becomes a limitation just for growth at very high rates (simulating batch conditions) but not for chemostat simulations with microbial models.

BenjaSanchez:
Something to address here brought up by @sulheim:

> parameter file can also provide the paths for the required database-files, cultivation data etc.

edkerk:
This will be completely revamped with GECKO3, the discussion here is obsolete.

PR #104, opened 2020-06-15
refactor: databases

mihai-sysbio:
We are working towards and automated pipeline in the `ecModels` repository. For that to work fine, it would really simplify things if GECKO as tool would be "clean", ie the tool itself would be separate from its application on the `yeast-GEM`.
I vaguely remember a discussion with @IVANDOMENZAIN about having something like a `article-year` branch that would contain such applications and associated figures, while allowing the `master` and `devel` to be tidy without having to carry forward the reproducible components throughout future updates of the code.
This PR only contains the database cleanup, but I think it would make sense to first setup the reproducible branches before anything.


BenjaSanchez:
@mihai-sysbio the main 2 reasons for these files to be here are:

1. For other people that implement GECKO for their own models, as they serve as templates that they can check to see how their data should look like.
2. For easy (manual) testing, as each time we modify the code we run the full pipeline on the yeast model to ensure no unexpected changes are introduced.

Therefore, I don't think it's a good idea to remove these files. The branched solution you propose sounds a bit overly-complicated to me, as we would have to keep an additional branch up to date to achieve the 2 objectives presented above, and it could confuse users that are not familiar with Github.

Why not do the cleanup from the ecModels repo side? Let me know what you think, and also @IVANDOMENZAIN @edkerk feel free to give your opinion on this.

mihai-sysbio:
Thanks @BenjaSanchez - allow me to check my understanding:
1. People should be able to implement GECKO for their own models, therefore GECKO should come with at least one example implementation.
2. GECKO should be tested, and the chosen test model is yeast.

I _fully_ support both of the above. However, I would also add:

3. GECKO could aim to be packaged in a way that is model agnostic.
This is a bit related to 1, in the sense that the barrier of entry for newcomers should be as low as possible. For `ecModels` the implication would be that it would make it more straightforward, not having to run a script that deletes model specific files.  

I can see how a different branch per article could be tedious. At the same time, I value two advantages it brings. First, it allows bug fixes for those specific versions for a very long time in the future (post publication) without affecting the master branch. Secondly, it reduces the burden of backwards compatibility by essentially freezing the publication branches. But there are likely better ways.

Regarding local manual testing, one way I've seen this done is to not commit the test environment to the repository, so that each developer can have their own desired test. Alternatively, as in the [automated ecModels fork](https://github.com/mihai-sysbio/ecModels/tree/feat/auto-run/ecYeastGEM) , the `databases` folder was extracted out and provided as what I hope is a more obvious model dependency, instead of what could be misinterpreted as a GECKO dependency.

edkerk:
I tend to agree with @mihai-sysbio that GECKO would ideally be distributed in a way that is model agnostic. Significant progress has been made (see e.g. #51, #55, #76), but there is still room for improvement (to remain on-topic, I'll not discuss this further here). To my opinion, providing model-specific files are part of that discussion.

Regarding the points raised by @BenjaSanchez, this can also be solved by:
1. Having an example **script** for model generation and proteomics integration. [Currently there are `enhanceGEM` and `generate_protModels`, but to me this should be an example script where e.g. ecYeast-GEM is constructed and a proteomics dataset integrated. Might seem like a trivial difference, but I think it is clearer to the user to present it as a script and not a function. Also; `generate_protModels` has nested functions, loops for making multiple models, hard-coded relative locations for files, etc.. If I use GECKO, I don't want to have to change functions, I just want to change one script where different parameters are given to functions.] As part of this reconstruction script, the relevant branch (and even commit) of ecModels is cloned to receive the model specific files (could also be `websave` if you want to prevent the requirement of git). Add the ecModels folder in `.gitignore`.
2. Have a testing script (`testing/constructEcYeastGEM8.1.3.m`) that again clones the relevant commit from ecModels and writes some relevant non-binary files (what is now in the models folder) in the `testing` folder?

Regarding the discussion of publication-branches, I assume these will not be maintained unless necessary (and then only bug-fixes). But these branches are only for GECKO publications (where the software is updated), and not for publications that just use GECKO? Because shouldn't the latter just have its own repository, where that repo hosts a script that clones the exact commit for a particular GECKO version, or even has GECKO as submodule fixed at a particular commit?

BenjaSanchez:
Thanks for your suggestions! I agree with both of you on what ideally GECKO should be (model agnostic and easy to implement for any organism), but a bigger issue for achieving this is that many scripts are currently organism-specific (@edkerk has mentioned this) and therefore have to be adapted when converting the model (see the [README](https://github.com/SysBioChalmers/GECKO#usage) file). So then removing the data files in this PR should IMO only come together with all of those scripts made generic (hence achieving true agnosticism) + moving all yeast files elsewhere, which is a significant refactoring effort.

Answering to some other points:

> For `ecModels` the implication would be that it would make it more straightforward, not having to run a script that deletes model specific files

This is in comparison a minor effort to implement, so for having `ecModels` working soon it's probably easier to implement that script for now, which we can later delete.

> `article-year` branches

Just to clarify: The model that we use for testing in `master` is not the model from the publication, it's just a given yeastGEM version that we chose for consistency at some point. In terms of publications (so far just the one and soon a second one), I think a **release** is probably enough: As publications are static, even if there is a bug in the code, the code associated to that publication should keep it, to guarantee reproducibility with e.g. paper figures. It's worthy to distinguish between two types of "bugs":

1. Scientific bugs: Either we got the wrong biology and/or a mathematical part in the code does something unintended. Should not be corrected in the release version to guarantee reproducibility (but should be corrected in master).
2. Dependency bugs: A specific software that GECKO relies on updates/deprecates a function. In the 3 years that GECKO 1.0 came out, no changes have been needed to be made to that version (as Matlab tightly controls their environment). If something would be needed in the future due to one of these bugs (e.g. a certain Matlab version is no longer supported), we could indeed create a branch from that specific commit and re-assign the tag/release. But I doubt that would be of interest to any user, as they would probably be working with a more updated GECKO version.

@edkerk I believe this how RAVEN handles it?

> Have a testing script (`testing/constructEcYeastGEM8.1.3.m`) that again clones the relevant commit from ecModels and writes some relevant non-binary files (what is now in the models folder) in the testing

I think this would be a good approach: Keep everything yeast-specific (data files, input GEM, scripts, output ecModel) not in a branch but in a `test` or `tutorial` folder, so it's evident for the user how to use GECKO. That should also be included in this major refactoring effort.

> But these branches are only for GECKO publications (where the software is updated), and not for publications that just use GECKO?

Correct, any publication that uses GECKO should have their own repo.

edkerk:
Thank you @BenjaSanchez for the clear arguments. I agree with you that the first step would be to refactor all functions to be model agnostic, and `ecModels` to have all required species-specific files, and only in a last step delete the yeast-specific components of GECKO (unless these are kept in `tutorial`, as proposed). Perhaps we should raise an issue (or repurpose #51/#55) with a list of refactoring to be done, where removal of these files is at the bottom?

Now I better understand the idea of `article-year` branches, and indeed *releases* are probably sufficient for this purpose. The README can also include links to the releases in relation to the various GECKO publications. In **RAVEN** we have never updated an older release, we always make changes to `devel`, push to `master` and make a new sequential release. Also for GECKO I see no reason to maintain older releases. In the rare situation that this would be required this could be accomplished ad-hoc.

I like your idea of `tutorial` instead of `testing`, it would make more clear that the yeast-related scripts and datafiles are useful as a guidance on how to use GECKO.

mihai-sysbio:
To follow up the `tutorial` thread, instead of having parameters for all the files, as suggested in #55, I would rather go for an "apply dependencies from folder" script, where the default value for the folder would be "../tutorial". That's my personal taste though.
To summarize, _releases_ is the way forward to associate with publications.

It sounds like the plan for now is to duplicate the database files in `ecModels`, and as part of the automation to delete them and add new model specific ones before running GECKO.  This PR can imho be closed, thank you both for the discussion.

BenjaSanchez:
@mihai-sysbio @edkerk thank you both for bringing up this discussion and all the valuable points :)

PR #107, opened 2020-07-27
fix: match only growth rxn id

BenjaSanchez:
In the utility `generate_protModels.m`, only the growth rxn's LB should be set, not also the biomass pseudo-rxn's, as otherwise it will double-constrain the model and sometimes create undesired behavior (e.g. unfeasible simulations)


IVANDOMENZAIN (approved):
Thanks for pointing this out and fixing it @BenjaSanchez, I agree with the change and have already tested everything locally without any problems :)

Issue #102, opened 2020-06-04
fix: switch UBs from Inf to +1000?


BenjaSanchez:
I've noticed that ecModels give an `unbounded` error in cobrapy (but also sometimes with cobratoolbox/RAVEN), because most upper bounds are defined as `Inf`. @IVANDOMENZAIN is there any specific reason that we chose `Inf` instead of `+1000` for the default upper bound? If none then I can change it throughout the toolbox, to guarantee that ecModels work out of the box without needing any specific edits.

IVANDOMENZAIN:
@BenjaSanchez No I don't remember any specific reason for that, I agree on introducing such fix :)

PR #109, opened 2020-10-01
Fix: convert +Inf to +1000 throughout the toolbox

BenjaSanchez:
After discussion in #102, this PR changes all +Inf bounds to +1000. Also updated all models in the tool. @IVANDOMENZAIN let me know if I forgot to update something.


Midnighter:
Sorry, I'm a bit late to the party. Doesn't the unbounded error reveal an underlying problem and adding a finite upper bound just masks this? I say this without really knowing in which cases this occurs and what the reasons are.

BenjaSanchez:
@Midnighter Right now I can't remember exactly what triggered this convention change, but a couple examples where I think `+1000` is preferable over `+Inf`:
* As ecModels have an irreversible structure (reaction A <-> B translates to 1 reaction for each direction), internal cycles that can take any value are quite common, and I believe made the model turn "unbounded" a couple times (can't remember if with cobrapy or cobratoolbox).
* When running FVA or any random sampler for an ecModel, it's easier to record fully unconstrained fluxes, as there's no need of building try/except blocks.

Considering that `+1000` is pretty standard in the field, I don't think the change would be a problem?

Midnighter:
> Considering that +1000 is pretty standard in the field, I don't think the change would be a problem?

No, it's not a problem but it's one of those cases where I'd say it's actually a bad habit :slightly_smiling_face: Finding better solutions can be difficult, though, so don't let me stop you!

IVANDOMENZAIN (approved):
Everything looks good, I didn't find any other function related to the imposed bounds requiring any adjustments. Ran the pipeline and my output files match with the uploaded ones after having updated COBRA. Ready yo merge!

Issue #112, opened 2020-10-05
fix: correct retrieval of original model version


IVANDOMENZAIN:
It is important to save `git` readable ecModel output files indicating the correct version of the corresponding source**GEM** in order to facilitate the tracking of changes across different model versions. The last saving steps in the `enhanceGEM.m` pipeline rely on the **`RAVEN toolbox`** for incorporation of the ecModel version as a string into the `model_id` field.

Such version string can be provided as an input for the `enhanceGEM` function or generated as an output by `preprocessModel`. Inside this function, the version can be parsed from the original GEM `id` field, or typed by the user if this field is missing. 

The current solutions seem to be robust as it provides different alternatives to ensure that a version is included in the final ecModel files, however the whole GECKO pipeline stores and deals with this version string in a variable called `version` (in all of the functions in which it is used), this term coincides with the name of a MATLAB built-in function that returns the installed MATLAB version. The updated ecModel output files in #111 highlight this issue. 

The use of MATLAB native or reserved terms should be avoided across the toolbox in order to prevent this kind of ambiguities to happen.

BenjaSanchez:
@IVANDOMENZAIN I can't seem to reproduce this behavior, as in my setup the proper version gets stored (see latest commit in https://github.com/SysBioChalmers/GECKO/pull/111). Perhaps it's a Mac issue? In any case, it's probably safest to do the change, considering that "version" is a pretty generic term :)

PR #113, opened 2020-10-06
Fix/ec model versioning

IVANDOMENZAIN:
This PR addresses #112: in order to provide a correct versioning system for ecModel git-readable files, the use of variables called `version` across the toolbox has been substituted by a non-ambiguous variable name in all of the functions that required it (`preprocessModel.m` and `enhanceGEM.m`). Git-readable ecModels files show the correct version of yeastGEM now.


BenjaSanchez (approved):
working well over here!

PR #119, opened 2020-12-10
Feat/add fseof utilities

IVANDOMENZAIN:

A series of scripts for running the [FSEOF algorithm](https://pubmed.ncbi.nlm.nih.gov/20348305/) suitable for ecModels has been added. The main function `run_ecFSEOF` runs the algorithm for a provided rxn target and using a specified carbon source. Lower and upper values for a suboptimal biomass yield interval should be provided by the user. Outputs are given as tables of rxn and genes targets (for knockdown, knockout and overexpression), additionally `.txt` can also be produced if file names are provided. 


geckomat/utilities/ecFSEOF/run_ecFSEOF.m
```diff
@@ -0,0 +1,69 @@
...
+model.grRules        = grRules;
+model.rxnGeneMat     = rxnGeneMat;
+%run FSEOF analysis
+results = ecFlux_scanning(model,rxnTarget,cSource,alphaV,1E-4,true);
+%Create gene table:
+results.geneTable      = cell(length(results.genes),3);
+results.geneTable(:,1) = results.genes;
+results.geneTable(:,2) = results.geneNames;
+results.geneTable(:,3) = num2cell(results.k_genes);
+%Create rxns table (exclude enzyme usage reactions):
```
BenjaSanchez, line 46: Just a thought here: Why exclude enzyme usage reactions from the analysis? Considering that those reactions translate 1-1 to genes, could it be an alternative way of getting gene candidates? I ask because in other FSEOF analysis I've conducted on ecModels (for instance [this](https://github.com/BenjaSanchez/notebooks/blob/master/post-ec-model-biotin/notebook.ipynb) one), I have **_only_** looked at the enzyme usage reactions.

Also, I noticed that in the case study I ran for `ecYeast` when testing this branch, there were more enzyme usage reactions (562-304=258) than genes (126), have you seen this as well in your tests?

geckomat/utilities/ecFSEOF/run_ecFSEOF.m
```diff
@@ -0,0 +1,69 @@
...
+% 		alphaLims (vector) Minimum and maximum biomass yield [gDw/mmol Csource] 
+%				  for enforced objective limits
+%       Nsteps    (integer) Number of steps for suboptimal objective 
+%                 in FSEOF
+%		file1 	  (string) opt, file name for an output .txt tab-separated 
+% 				  file for the results at the genes level
+%		file2 	  (string) opt, file name for an output .txt tab-separated 
+% 				  file for the results at the reactions level		
+%
+% Usage: run_ecFSEOF(ecModel,rxnTarget,cSource,alphaLims,Nsteps)
```
BenjaSanchez, line 22: Although the inputs are all clearly explained, the output structure is not, so I would recommend including a summary of what does each field in the `results` structure represent.

geckomat/utilities/ecFSEOF/run_ecFSEOF.m
```diff
@@ -0,0 +1,69 @@
...
+results.geneTable(:,3) = num2cell(results.k_genes);
+%Create rxns table (exclude enzyme usage reactions):
+enzIndxs               = find(~startsWith(results.rxns(:,1),'draw_prot_'));
+results.k_rxns         = results.k_rxns(enzIndxs);
+results.rxnsTable      = cell(length(results.k_rxns),5);
+results.rxnsTable(:,1) = results.rxns(enzIndxs,1);
+results.rxnsTable(:,2) = results.rxns(enzIndxs,2);
+results.rxnsTable(:,3) = num2cell(results.k_rxns);
+results.rxnsTable(:,4) = results.rxns(enzIndxs,3);
+results.rxnsTable(:,5) = results.rxns(enzIndxs,4);
```
BenjaSanchez, line 54: If you are creating the summaries `rxnsTable` and `geneTable`, then I would suggest removing from the results structure `k_rxns`, `rxns`, `genes`, `geneNames` and `k_genes`, as they are all redundant information.

Also, consider filtering `v_matrix` and `k_matrix` as well, to avoid having reaction fields of different sizes:

image

BenjaSanchez (commented):
Nice work! See some minor comments below, should be ready to merge after addressing them :)

geckomat/utilities/simulateGrowth.m
```diff
@@ -0,0 +1,86 @@
...
+end
+%Fix a unit main carbon source uptake
+cSource_indx           = find(strcmpi(model.rxnNames,C_source));
+model.lb(cSource_indx) = (1-tol);
+model.ub(cSource_indx) = (1+tol);
+%Position of target rxn:
+posP      = strcmpi(model.rxns,target);
+growthPos = find(model.c);
+%Max growth:
+sol = solveLP(model);
```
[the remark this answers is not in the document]
IVANDOMENZAIN: Good comment, addressed in 5929e56

geckomat/utilities/ecFSEOF/run_ecFSEOF.m
```diff
@@ -0,0 +1,69 @@
...
+% run_ecFSEOF
+%
+% 	Function that runs Flux-scanning with Enforced Objective Function
+% 	for a specified production target.
+%   
+%		model     (struct) ecModel with total protein pool constraint
+%		rxnTarget (string) Rxn ID for the production target reaction, 
+% 				  a exchange reaction is recommended.
+%		cSource	  (string) Rxn name for the main carbon source uptake 
+%			      reaction
```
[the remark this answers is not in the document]
IVANDOMENZAIN: Fixed in 44b0368

geckomat/utilities/ecFSEOF/run_ecFSEOF.m
IVANDOMENZAIN: I agree. Both suggestions have been addressed in: fc0df92 and  291640f

geckomat/utilities/ecFSEOF/run_ecFSEOF.m
IVANDOMENZAIN: > Just a thought here: Why exclude enzyme usage reactions from the analysis? Considering that those reactions translate 1-1 to genes, could it be an alternative way of getting gene candidates? I ask because in other FSEOF analysis I've conducted on ecModels (for instance [this](https://github.com/BenjaSanchez/notebooks/blob/master/post-ec-model-biotin/notebook.ipynb) one), I have **_only_** looked at the enzyme usage reactions.
> 
> Also, I noticed that in the case study I ran for `ecYeast` when testing this branch, there were more enzyme usage reactions (562-304=258) than genes (126), have you seen this as well in your tests?

Very interesting comment @BenjaSanchez! I've addressed this (partially) in fbad8c6. Short summary:

It is strange that the algorithm returns more `draw_prot_XXXX` reactions than genes in the final `geneTable`. I checked the code and remembered that there is a step that removes all of those gene targets considered as "unchanged" (gene targets with a  `mean(K_score) < K_score < 1`), which makes sense in order to just report the most relevant results. Therefore the mismatch in the numbers was caused by those remaining `draw_prot_XXXX` rxns in the `rxnTable` associated to genes that are removed in the aforementioned step. As a fix, commit fbad8c6 updates the rxn-related fields, making sure that any rxns related to the removed "unchanged" genes are also removed from the results.

IVANDOMENZAIN: > Just a thought here: Why exclude enzyme usage reactions from the analysis? Considering that those reactions translate 1-1 to genes, could it be an alternative way of getting gene candidates? I ask because in other FSEOF analysis I've conducted on ecModels (for instance [this](https://github.com/BenjaSanchez/notebooks/blob/master/post-ec-model-biotin/notebook.ipynb) one), I have **_only_** looked at the enzyme usage reactions.
> 
> Also, I noticed that in the case study I ran for `ecYeast` when testing this branch, there were more enzyme usage reactions (562-304=258) than genes (126), have you seen this as well in your tests?

As a follow up to the initial question (@BenjaSanchez): The enzyme usage reactions are not excluded from the analysis, they are just removed from the final list of rxns results (trying to just show metabolic reactions for simplicity). However, these reactions are accounted for in previous steps (inside of the `ecFlux_scanning` function), therefore the genes associated to them should be included in the `genesTable` results. The intention is to keep results separated by rxns and genes, both having different fields however their connection is described in the column `grRules` in the `rxnsTable`.

Even though the`draw_prot_XXXX` reactions provide a 1-1 relation to genes, there is risk of ignoring potential gene targets that are not directly associated to any enzyme in the ecModel structure. This problem is not so acute for `ecYeastGEM` in which the enzyme coverage is high, but it might be relevant for industrially relevant species whose corresponding ecModels present a lower enzyme coverage (e.g. *Y. lipolytica* and *K. marxianus*).

IVANDOMENZAIN:
@BenjaSanchez Finally, in order to reduce the amount of redundant gene and reaction targets returned by the algorithm, commit eede369 discards the `arm_rxns` that are upregulated (consistently increasing K_scores) prior to map their respective coding genes. 

If such reactions are kept, then all their associated isoenzymes are included in the genes targets and the most common scenario is that just the most efficient isoenzymes carry flux for a given objective. Therefore, removing upregulated `arm_rxns` at the very beginning of the method ensures that just the flux-carrier isoenzymes (most efficient ones) are suggested as targets for over expression in the final `genesTable`.

The same rationale cannot be directly applied to downregulated `arm_rxns`, as for some cases it might be relevant to remove **all** isoenzymes associated to reactions that do not contribute to maximize product formation.

geckomat/utilities/ecFSEOF/run_ecFSEOF.m
BenjaSanchez: Roger that! all good then, thanks for looking into it.

BenjaSanchez (approved):
PR is ready to merge now, @IVANDOMENZAIN thanks for incorporating the suggestions.

IVANDOMENZAIN:
> PR is ready to merge now, @IVANDOMENZAIN thanks for incorporating the suggestions.

Thanks for a thorough and kind review!

PR #127, opened 2021-01-11
Fix/updates to getSubset_ecModel

JonathanRob:
A few changes were made to the `getSubset_ecModel` function to make sure that it runs as expected. I've summarized the changes below with some explanations.

**Function header**: 1b9534b
I just noticed some of the variable names hadn't been updated in the function header, so I corrected those.

**Use default UB**: 11b53a5
I often use a default upper/lower bound of `+/-Inf` rather than `+/-1000` when working with ecGEMs due to numerical solver issues. The function now uses the value from `.annotation.defaultUB` if the field exists in the ecGEM structure.

**Perform exact pattern matching for reactions**: 5c0f5c7; refactored: 6dd91f0
Previously, the code was searching for ecGEM reactions whose names contained any of the `smallGEM` reactions. For example, for `smallGEM` reaction `HMR_1000`, it would find its "EC" versions such as `arm_HMR_1000`, `HMR_1000_REV`, `HMR_1000No1`, etc., depending on the grRule associated with `HMR_1000`.

The problem with this approach is that some reaction names are subsets of others, which will result in erroneous matching. For example, if we were looking for all reactions containing `HMR_1000`, we would also match `HMR_10001`, `HMR_10002`, `HMR_10003`, etc., which are completely different reactions.

To address this, the function identifies the original GEM reaction ID associated with each ecGEM reaction ID by removing the prefix `arm_` and/or suffixes `_REV` or `No#` (where `#` is an integer(s)). For example, `arm_HMR_1000_REVNo14` becomes `HMR_1000`. These trimmed IDs can then be evaulated for exact matches to reaction IDs in `smallGEM` to determine which ecGEM reactions should be retained.

If there are any other modification types that should be matched, then let me know so we can incorporate it into the search pattern.

**More enzyme-related reaction types**: 5e15cd7
I noticed that the function finds reactions matching `draw_prot_XXX` (where XXX is an enzyme ID) and `prot_pool_exchange`, but doesn't match reactions of the form `prot_XXX_exchange` which I see in some ecGEMs. I've updated it so it now detects reactions like `prot_XXX_exchange`, but maybe this is incorrect?

**Exact matching of enzyme pseudometabolites**: ed37ac8
Minor change to match exact pseudometabolite names, such as `prot_XXX`, to avoid accidental substring matching. This is probably not necessary since I doubt the protein IDs have the same problem as the reaction names above, but good to be safe.

**Missing structure field**: 2825b0d
When I was testing the function, it threw an error because the ecGEM I used didn't have the `concs` field. I modified the last part of the code to only edit field names that were present in the ecGEM to avoid this error in the future.

### DIFF
--- a/geckomat/utilities/getSubset_ecModel.m
+++ b/geckomat/utilities/getSubset_ecModel.m
@@ -1,5 +1,5 @@
 function small_ecModel = getSubset_ecModel(smallGEM,big_ecModel)
-%getsmall_ecModel
+%getSubset_ecModel
 %  
 % Generate a context-specific ecModel (strain/cell-line/tissue) by mapping 
 % all components in an original context-specific GEM to a general ecModel
@@ -10,9 +10,9 @@ function small_ecModel = getSubset_ecModel(smallGEM,big_ecModel)
 %   smallGEM       Reduced model (subset of the general model) for a
 %                  specific strain (microbes) or cell-line/tissues (mammals)
 % 
-%   ecModel        Enzyme-constrained version of the context-specific model
+%   small_ecModel  Enzyme-constrained version of the context-specific model
 % 
-% usage: small_ecModel = getsmall_ecModel(smallGEM,big_ecModel)
+% usage: small_ecModel = getSubset_ecModel(smallGEM,big_ecModel)
 % 
 
 %If smallGEM is in COBRA format convert to RAVEN-type
--- a/geckomat/utilities/getSubset_ecModel.m
+++ b/geckomat/utilities/getSubset_ecModel.m
@@ -39,8 +39,14 @@ small_ecModel = removeGenes(big_ecModel,toRemove,true,true,true);
 %that are not represented in the context-specific GEM, mostly non gene-associated reactions)
 originalRxns = smallGEM.rxns;
 toKeep       = [];
-for rxn = originalRxns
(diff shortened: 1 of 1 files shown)


JonathanRob:
Amazing - with @mihai-sysbio's suggestion and the [recent changes](https://github.com/SysBioChalmers/RAVEN/pull/323) made to the RAVEN `removeGenes` function, I was able to subset ecHumanGEM in under 30 sec.

IVANDOMENZAIN:
Thanks @JonathanRob and @mihai-sysbio for your valuable input and work on this PR, really good job! same results here, the computation time has been drastically reduced.

JonathanRob:
Interesting that you have seen more issues with `Inf` as bounds rather than `+/-1000`. I often run into infeasibility errors when using `+/-1000`, which are resolved when changing the bounds to `Inf`. But I suppose this problem is likely very model- and solver-dependent.

JonathanRob:
@IVANDOMENZAIN Ah, ok, I figured that there was a reason you excluded these `prot_XXXX_exchange` reactions in your approach. Yes, I agree that your solution to include a warning about potentially over-constraining the model would be a good approach for now - if we realize that they really should not be included, then we can always turn it into an error or revert to your original approach.

PR #134, opened 2021-02-05
feat: specify carbon source with fermentation data

edkerk:
When generating multiple proteomics-constrained ec-models with different carbon sources, this currently would require different `getModelParameters.m` and `fermentationData.txt` files. Instead, in this PR:
- the carbon source exchange reaction name is included in `fermentationData.txt`
- `load_Prot_Ferm_Data.m` places this information in `fermParameters.c_source`
- `generate_ProtModels.m` uses the condition specific `fermParameters.c_source` instead of the information from `getModelParameters`. (It does not replace `parameters.c_source` in `getModelParameters`, as the carbon source mentioned here is specific for each proteomics dataset)

- some duplicated code (see line 25-27 and 37-39 in old file) was removed.


edkerk:
Last thing to fix here, although not just related to using different carbon sources, is that if `generate_protModels` runs for multiple conditions is overwrites the model file in every iteration, in location `models/prot_constrained/NAME/NAME.xml`. If can think of three solutions, of which I now implemented the first option (commit bf8aaea76ad30df229fda2669d6ffcbfc31cc1af):

1. Rename the file in each iteration to `models/prot_constrained/NAME/NAME_CONDITION.xml`
2. Alternatively save the file and the additional tables in `models/prot_constrained/NAME_CONDITION/NAME.xml` and comparable.
3. Alternatively as 2, but only if the number of conditions > 1, otherwise save as 
`models/prot_constrained/NAME/NAME.xml` (as the original code).

(Also did a minor change of `saveECmodel` when toolbox 'RAVEN' is specified (commit 6165161), it now writes all model files in the same folder, same behaviour as COBRA),

edkerk:
The changes intended here have been superseded by various developments as part of GECKO3.

PR #135, opened 2021-02-11
feat: allow non-default GAM parameters

### LINKED ISSUE #133: feat: remove hard-coded params fitGAM
Some hard-coded parameters could be moved out of the `fitGAM` function, to make the code more versatile, and so no custom function edits would be required if a model would require non-default parameter values.

1. the scanned GAM range (`GAM = 20:5:100`) could be mentioned in `getModelParameters and set to these values by default?

2.  the title and legend of the figure generated contain the name of the carbon source
    2.1  `fitGAM` uses `chemostatData.tsv`, take the carbon source name from there?
    2.2  alternatively, specify carbon source / growth medium name in `getModelParameters`?

edkerk:
Fixes #133:
- `getModelParameters` contains the optional `parameters.maxGAM`, which can be used to specify the scanned GAM range (GAM = 20:5:100) in `fitGAM`. If not specified, a maximum GAM of 100 is used by default, as was already the case.
- title and legend of the figure generated by `fitGAM` takes the carbon source from `chemostatData.tsv`

### DIFF
--- a/geckomat/getModelParameters.m
+++ b/geckomat/getModelParameters.m
@@ -16,6 +16,10 @@ parameters.Ptot = 0.5;      %Assumed constant
 %Minimum growth rate the model should grow at [1/h]
 parameters.gR_exp = 0.41;     %[g/gDw h] 
 
+%Set GAM parameters (optional)
+%parameters.GAM = 100; %Use fixed GAM instead of fitting by fitGAM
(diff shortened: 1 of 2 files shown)


edkerk:
The changes intended here have been superseded by various developments as part of GECKO3

PR #146, opened 2021-03-05
Fix/pr suggestions

IVANDOMENZAIN:
I'm not fully convinced by a suggestion in PR #130, namely, the rescaling of `Ptot` after the first flexibilization of enzyme abundances is performed (prior to enter the `constrainEnzymes` function in `generate_protModels`). Ideally, `Ptot` should be quantified independently by a specific experimental assay, such as the [Lowry protocol](https://en.wikipedia.org/wiki/Lowry_protein_assay). In my experience analyzing data on multiple conditions for three different yeast species, this is a measurement with a very low associated standard deviation across technical and biological replicates. In contrast, it is well known that the available methods for absolute proteomics quantification display very high levels of both technical and biological variability (that's for filtering out some values prior to imposing the constraints). 

My concern here is that we are flexibilizing `Ptot`, an independent and quite robust measurement, according to the needed flexibilizations on the individual enzyme abundances, which are intrinsically highly variable. Moreover, we shouldn't forget that flexibilizations in individual protein levels might be caused due to wrong abundance estimations or by erroneous **Kcats**, which come from myriads of independent studies with non-standardized conditions.

I'm aware that `Ptot` flexibilization is one of the key features in PR #130 and it might enable the obtention of feasible models for highly protein-constrained conditions (low cell protein content or at maximum growth rates), but this step might be introducing some additional protein mass ad-hoc into our models. The following changes were introduced trying to provide a comprehensive sequence of steps to constrain `Ppool` for a wide variety of situations.

- Flexibilize all saturated enzymes when no individual enzyme limitations are found (3b1e2b6)
- Revert flexibilization of `Ptot` prior to obtention of a protein constrained model (1df2821)
- Condition-specific `f factor` is just used when paxDB file is not available or when there's no overlap between the model genes and those in the paxDB file (paxDB for another organism) (655b7a4 & 1d7979b), for all other cases the original `f factor` calculation is performed (based on paxDB file). This `revert` is derived from the ongoing [discussion in PR #130](https://github.com/SysBioChalmers/GECKO/pull/130#issuecomment-782971512). 
- Restructure `updateProtPool` function (9a38ce5)
- Improve output messages regarding calculation of `protein pool UB`.

Overall, these changes suggest the following sequence of steps for overconstrained models with proteomics
1. Full saturation of enzymes in the protein pool (sigma=1) is allowed prior to `flexibilizeProteins` step. This could contribute to reduce the total mass of flexibilized proteins.
2. Iterative flexibilization of top-limiting enzymes until the model reaches the experimental growth rate. If no individual limitations are found then all fully-saturated enzymes are flexibilized by 1% (limiting enzyme complexes).
3. The flexibilized enzyme mass is added to `Pmeas` for recalculation of a protein pool UB.
4. Recalculation of the protein pool UB:
4.1. First attempt: the initially calculated `f factor` and the default `sigma` are used. 
4.2. If the resulting model is unfeasible, then `sigma` is refitted according to minimal protein requirements. 
4.3. If the refitted `sigma` is higher than one, then `sigma=1` is fixed and the `f factor` is refitted according to minimal protein requirements. 
4.4. In case that the minimum protein requirements yield an `f factor` higher than one, then a model with an unconstrained pool and a warning message are returned.


IVANDOMENZAIN:
@edkerk My apologies for taking so long with PR #130, I hope that this can contribute to the discussion. Could you try to test this suggestions with your low protein cases?

edkerk:
I'll test it, thank you for your thorough review!

IVANDOMENZAIN:
@edkerk some protein content data generated by the CHASSY project using the Lowry protocol.

Total protein content of *S. cerevisiae* cells growing at 0.1h-1 [g protein/gDW]
Screen Shot 2021-03-05 at 20 31 37
Total protein content of *K. marxianus* cells growing at 0.1h-1 [g protein/gDW]
Screen Shot 2021-03-05 at 20 32 58

geckomat/utilities/integrate_proteomics/generate_protModels.m
```diff
@@ -88,9 +85,10 @@ function generate_protModels(ecModel,grouping,name,ecModel_batch)
     cd ../kcat_sensitivity_analysis
     ecModelP  = changeMedia_batch(ecModel,c_source);
     tempModel = changeMedia_batch(ecModel_batch,c_source);
-    cd ../limit_proteins
-    %Calculate f-factor, after having filtered data. 
-    f = measureAbundance(ecModel.enzymes,pIDs,abundances);
+    cd ../limit_proteins  
+    if isempty(f) | f==0
```
edkerk, line 89: Even if `f = []` in line 52, when `i = 2`, `f` is no longer empty. Not sure how to fix, resetting `f` at the end of the `i`-loop would also overwrite any `f` that could have been set in line 48.

geckomat/limit_proteins/constrainEnzymes.m
```diff
@@ -90,16 +91,31 @@
...
+    else
+        %Expected total enzyme concentration
+        enzymeConc=Ptot*f;
+        %Non-measured part will be pooled
+        PpoolMass = enzymeConc-Pmeasured;
+        %If not all enzymes are measured and there's no remaining mass for
+        %the protein pool, then assume f=1
+        if sum(measured)<length(model.enzymes) & PpoolMass<=0
+        	warning('The measured mass of protein exceeds the remaining amount of unmeasured protein in the model, assuming f=1 (all protein mass is available for metabolic enzymes) in order to get a feasible model')
+            PpoolMass = Ptot - Pmeasured;
```
edkerk, line 105: If `f=1` is assumed, then it should also be set, otherwise `updateProtPool` in lines 134 and 137 will fail.
```suggestion
            PpoolMass = Ptot - Pmeasured;
            f=1;
```

edkerk (changes requested):
It works (after including changes from #147, but that seems not related to the issue here), and with my suggested code changes. Still one of the models gets `Pmeasured > Ptot*f`, but now it very nicely leaves the protein pool exchange UB at `Ptot*1*sigma`. In that case, it would probably make sense to minimize the protein pool exchange UB after model generation (basically minimizing `f`, assuming `sigma` is unchanged), as the current UB is now _very_ lenient. But that should probably just be up to the user to do themselves.

PR #150, opened 2021-07-02
opt: Optimization of getEnzymeCodes and MatchKcats. 

johan-gson:
…ecko Light. The findInDBOld is still needed for KEGG matches, I haven't looked at that.

Some new optimizations were also introduced since there were other bottlenecks for yeast. I use containers.Map which may require a somewhat new matlab version.


johan-gson:
I had to add some additional optimizations for yeast. Check required matlab version for containers.Map, I use that. Also, I think it would be possible to optimize convertToEnzymeModel, that is pretty slow. The code is pretty similar to the GetMWandKCats in GeckoLight, the same optimizations could probably be done in that function. I also recommending creating lists of all mets/rxns to add, and add them in one call, instead of multiple calls. I tested that the old and new code produce the same kcat output, but some additional testing would be good.

johan-gson:
eh, I managed to close it accidentally

ckitti (approved):
Reviewed GECKO model generation with iJN1462 GEM

ckitti (approved):

mihai-sysbio:
I think this was merged prematurely:
- the target branch should not be `master` but `devel`
- there is a bunch of code that is simply commented out (either remove it or use it)
- there is no associated issue that describes what the problem is so that it can be discussed
- the PR does not summarize the implementation

I would strongly advise reverting this merge to address the points above.

Issue #152, opened 2021-11-02
Optimization of Gecko


johan-gson:
### Description of the new feature:
Introduction of some optimizations for Gecko regarding database handling. No functional changes are introduced.

**I hereby confirm that I have:**
- [x] Done this analysis in the `master` branch of the repository
- [x] Checked that a similar issue does not exist already

BenjaSanchez:
@johan-gson this issue lacks explanation about what "optimizations" are needed. In any case, it is not necessary to open an issue at the same time you are opening a PR that solves it (I believe #153?)

johan-gson:
This issue is not really needed, the info is given in the pull request.

johan-gson:
So, I was developing GeckoLight and discovered that running the first steps of Gecko, getEnzymeCodes and matchKcats, took like 20 minutes to run, which was very annoying for my experiments (and for other people I guess). After a quick discussion with Jens I spent two days optimizing the code, and for human it is now down to less than 2 minutes.

mihai-sysbio:
Interesting insight! I guess there weren't major changes in terms of memory usage?

johan-gson:
No, I don't think so, just a few indexes, which indexes in the order of 10000 items, so that is nothing.

mihai-sysbio:
This issue was resolved through PR #153

PR #153, opened 2021-11-02
Opt/optimize kcats processing attempt 2

johan-gson:
This is a fix of the previous pull request with the same name. That was accidentally merged to master, and was followed by a pull request to remove it again. This adds the code again, but to devel this time. Connected to issue #152

Cheewin has already reviewed the code (see the previous pull request). What is new is that I added an issue (#152) and removed some commented out code.

In short, there are many small optimization steps introduced in this PR. The most advanced ones rely on pregenerating indexes for faster search, which speeds things up considerably in the main loops. The details are not given here, see the code for more details


johan-gson:
I will make a third attempt, where I have branched from devel instead. I new pull request will soon come, I will close this one.

PR #154, opened 2021-11-02
Introduction of optimizations originally developed in Gecko Light.

johan-gson:
…e optimizations concern getEnzymeCodes and matchKcats. The findInDBOld is still needed for KEGG matches, I haven't looked at that.

Some new optimizations were also introduced since there were other bottlenecks for yeast. I use containers.Map which may require a somewhat new matlab version.

This is a continuation of a series of flawed pull requests, but I think this should be right. 

Cheewin has done most of the testing, I have run some simple tests, where it produced identical results as before.

In short, there are many small optimization steps introduced in this PR, originally developed in GeckoLight, that speeds up getEnzymeCodes and matchKcats with roughly a factor 10. The most advanced ones rely on pregenerating indexes for faster search, which speeds things up considerably in the main loops. I also introduced some additional optimizations not present in GeckoLight that turned out to be important for yeast. I used containers.Map which may require a somewhat new matlab version.


mihai-sysbio:
> This is a continuation of a series of flawed pull requests, but I think this should be right.

No worries, I'm glad to see this advancing.
 
> Cheewin has done most of the testing, I have run some simple tests, where it produced identical results as before.

It would be nice to hear more about the testing @ckitti (especially if done for different setups).

> In short, there are many small optimization steps introduced in this PR, originally developed in GeckoLight, that speeds up getEnzymeCodes and matchKcats with roughly a factor 10. The most advanced ones rely on pregenerating indexes for faster search, which speeds things up considerably in the main loops. I also introduced some additional optimizations not present in GeckoLight that turned out to be important for yeast.

That sounds amazing 🚀 

> I used containers.Map which may require a somewhat new matlab version.

Can you specify which?

johan-gson:
So, we investigated if the version would be a problem some months ago, and concluded it was not a large problem. I don't remember exactly which version, it is not super easy to figure it out, I spent some time on it last time.

mihai-sysbio:
> So, we investigated if the version would be a problem some months ago, and concluded it was not a large problem. I don't remember exactly which version, it is not super easy to figure it out, I spent some time on it last time.

According to the [documentation](https://se.mathworks.com/help/matlab/ref/containers.map.html), `containers.Map` has been "Introduced in R2008b". It's surprising that this would have created problems.

johan-gson:
Hmm, ok, no, this is what I used. Not much of a problem then.

edkerk:
FYI, RAVEN also uses `containers.Map` in various functions, and has done so "forever" (at least 2016), I've never heard of anybody running into problems with it.

Cheewin has constructed an ec-model with this PR and the previous code, and got identical models.

johan-gson:
Ok, I take it he will make some kind of review here, although it has already been reviewed, so I can finally merge this...

edkerk (approved):

PR #166, opened 2022-06-23
feat: makeEcModel and applyKcatConstraints

edkerk:

These functions define a new model structure where kcat values are stored and can be reapplied to the model at any time. This forms the basis with which we can refactor the code. This is not final, but forms a basis to start from.

The functions are in `geckomat/change_model/`, there are extensive explanations at the top of each function file.

Also introduced here:
- `findGECKOroot`: within functions you should not have to navigate to specific folders to run particular GECKO functions, instead GECKO should just be added to MATLAB path. But sometimes you still need to have the full path to a particular location, because this is where some input data is stored. This function will give you the path to GECKO root.
- Instead of a `mat` file with Uniprot data, there is a TSV file in `Databases`. `loadDatabases` should be able to download this file, but it's not working great yet, as you need to set the correct taxid first.

Further refactoring should be done by making branches from the `devel3` branch, and changes should be pushed back to `devel3`. Functions that have become obsolete because they have been replaced by another function can be indicated as such. They can still be kept for now, for easy reference to earlier code, but renaming them will at least make sure we all know that they will disappear.


ae-tafur:
Since sometimes you will need the full path because some functions call the path to the database files (e.g. chemostat or prot_abundance), you can add the path as a parameter in the getModelParameters.m file, and the functions that need it call it

edkerk:
I realize that this doesn't need to be thoroughly reviewed before merging. Instead, this can be the basis to work from and to modify. I'll therefore merge this PR in 1-2 days.

PR #168, opened 2022-07-07
feat: ModelAdapter pattern for organism specific parameters and customizations

johan-gson:
Added an adapter pattern that can enable us to use GECKO for different species without copying files. Right now the adapter is pretty small, but the idea is to add more functions in the base class when needed. The idea is then that an adapter is sent into GECKO when generating models, and that GECKO can ask the adapter for information that is accessed differently for each species/model. Examples include parameters, pathways to data, and which reactions are spontaneous. The code is not plugged in anywhere, so this is totally safe.

I also added an installation script, let's see what you think about that. To install (i.e., setup path): 
cd to GECKO root
GECKOInstaller.install
to uninstall (i.e., remove GECKO from path):
GECKOInstaller.uninstall

This is practical if you have several versions of GECKO on your disk and want to switch between them.

- [X] Tested my code with [all requirements](https://github.com/SysBioChalmers/GECKO) for running GECKO

### DIFF
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -28,8 +28,8 @@ classdef GECKOInstaller
 
         function newPaths = GetFilteredSubPaths(path_, filter_)
             pathSep = pathsep();
-			%Check that there are no semicolons in the path - that will cause 
-			%problems since that is the separator used to separate paths
+			%Check that there are no separators in the path - that will cause 
+            %problems since the separator is used to separate paths in a string
 			if contains(path_, pathSep)
 				error('The path in which GECKO resides may not contain semicolons for this installation to work!');
 			end
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -33,7 +33,7 @@ classdef GECKOInstaller
 				error('The path in which GECKO resides may not contain semicolons for this installation to work!');
 			end
             paths = genpath(path_);
-            splitPaths = strsplit(paths, ';');
+            splitPaths = strsplit(paths, GECKOInstaller.PATH_SEPARATOR);
             %remove the last, it is empty
             splitPaths = splitPaths(1,1:end-1);
             matches = regexp(splitPaths, filter_, 'match');
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -4,19 +4,17 @@ classdef GECKOInstaller
 %   Run GECKOInstaller.install to install (will set up the path in MATLAB)
 %   Run GECKOInstaller.uninstall to clear the path from MATLAB
 %   To install, you first need to cd to the GECKO root.
-%
-% Johan Gustafsson, 2022-07-07
-%
+
     methods (Static)
         function install
             sourceDir = fileparts(which(mfilename));
-            paths = GECKOInstaller.GetFilteredSubPaths(sourceDir, '.*\.git.*');
+            paths = GECKOInstaller.GetFilteredSubPaths(sourceDir, FILE_FILTER);
             addpath(paths);
             savepath;
         end
         function uninstall
             sourceDir = fileparts(which(mfilename));
-            paths = GECKOInstaller.GetFilteredSubPaths(sourceDir, '.*\.git.*');
+            paths = GECKOInstaller.GetFilteredSubPaths(sourceDir, FILE_FILTER);
             rmpath(paths);
             savepath;
         end
@@ -40,4 +38,8 @@ classdef GECKOInstaller
             newPaths = strcat(char(join(pathsLeft,';')),';');
         end
     end
+    
+    properties (Constant)
+      FILE_FILTER = '.*\.git.*';
+   end
 end
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -45,5 +45,6 @@ classdef GECKOInstaller
     
     properties (Constant)
       FILE_FILTER = '.*\.git.*';
+      PATH_SEPARATOR = ';';
    end
 end
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -27,24 +27,24 @@ classdef GECKOInstaller
 		end
 
         function newPaths = GetFilteredSubPaths(path_, filter_)
+            pathSep = pathsep();
 			%Check that there are no semicolons in the path - that will cause 
 			%problems since that is the separator used to separate paths
-			if contains(path_, ';')
+			if contains(path_, pathSep)
 				error('The path in which GECKO resides may not contain semicolons for this installation to work!');
 			end
             paths = genpath(path_);
-            splitPaths = strsplit(paths, GECKOInstaller.PATH_SEPARATOR);
+            splitPaths = strsplit(paths, pathSep);
             %remove the last, it is empty
             splitPaths = splitPaths(1,1:end-1);
             matches = regexp(splitPaths, filter_, 'match');
             okPaths = cellfun(@isempty, matches);
             pathsLeft = splitPaths(1,okPaths);
-            newPaths = strcat(char(join(pathsLeft, GECKOInstaller.PATH_SEPARATOR)), GECKOInstaller.PATH_SEPARATOR);
+            newPaths = strcat(char(join(pathsLeft, pathSep)), pathSep);
         end
     end
     
     properties (Constant)
       FILE_FILTER = '.*\.git.*';
-      PATH_SEPARATOR = ';';
    end
 end
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -39,7 +39,7 @@ classdef GECKOInstaller
             matches = regexp(splitPaths, filter_, 'match');
             okPaths = cellfun(@isempty, matches);
             pathsLeft = splitPaths(1,okPaths);
-            newPaths = strcat(char(join(pathsLeft,';')),';');
+            newPaths = strcat(char(join(pathsLeft, GECKOInstaller.PATH_SEPARATOR)), GECKOInstaller.PATH_SEPARATOR);
         end
     end
     
--- a/GECKOInstaller.m
+++ b/GECKOInstaller.m
@@ -1,0 +1,43 @@
+classdef GECKOInstaller
+% GECKOInstaller
+%   Support for installing and uninstalling
+%   Run GECKOInstaller.install to install (will set up the path in MATLAB)
+%   Run GECKOInstaller.uninstall to clear the path from MATLAB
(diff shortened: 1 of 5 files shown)


edkerk:
- GECKOInstaller is nice! This should be persistent, right? So after a restart, GECKO should still be in the path. Even people that only have one GECKO instance can then easily add it to the path.
- The adapter solution is very nice as it seems to be very fast as well. I'm not sure how you would now get access to the functions that it also defines?

GECKOInstaller.m
```diff
@@ -0,0 +1,43 @@
...
+%   Run GECKOInstaller.install to install (will set up the path in MATLAB)
+%   Run GECKOInstaller.uninstall to clear the path from MATLAB
+%   To install, you first need to cd to the GECKO root.
+%
+% Johan Gustafsson, 2022-07-07
+%
+    methods (Static)
+        function install
+            sourceDir = fileparts(which(mfilename));
+            paths = GECKOInstaller.GetFilteredSubPaths(sourceDir, '.*\.git.*');
```
mihai-sysbio, line 13: There is some duplicated code with regards to path definition (`.*\.git.*`), it would be much nicer to have paths defined only once.

GECKOInstaller.m
```diff
@@ -0,0 +1,43 @@
...
+        function path = getGECKOMainPath()
+			path = fileparts(which(mfilename));
+			path = strrep(path, '\', '/'); %get rid of backslashes in Windows
+			if ~endsWith(path, '/')
+				path = strcat(path,'/');
+			end
+		end
+
+        function newPaths = GetFilteredSubPaths(path_, filter_)
+            % Will fail if you have a directory containing ';'
```
mihai-sysbio, line 32: It looks like it would fail silently, which would then be tricky to debug. A more obvious failure would instead be easier to sort out.

mihai-sysbio (commented):
Very nice demonstration of how the adapters are supposed to work. I think @IVANDOMENZAIN might have an opinion here if these adapters should go in the `GECKO`repository or in `ecModels`. Also, I'm not sure why `getHumanGEMRootPath` needs to be defined outside the adapter.

GECKOInstaller.m
```diff
@@ -0,0 +1,43 @@
+classdef GECKOInstaller
+% GECKOInstaller
+%   Support for installing and uninstalling
+%   Run GECKOInstaller.install to install (will set up the path in MATLAB)
+%   Run GECKOInstaller.uninstall to clear the path from MATLAB
+%   To install, you first need to cd to the GECKO root.
+%
+% Johan Gustafsson, 2022-07-07
```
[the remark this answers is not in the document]
johan-gson: Fixed

GECKOInstaller.m
johan-gson: Fixed

GECKOInstaller.m
johan-gson: I'm thinking we can leave this as is - it is difficult to fix. The problem is that the matlab function I'm using to get all subdirectories, genpath, will fail in such cases - it returns the directories separated by ';', so such directories will be split in the returned string. There is no easy way to detect it. And, the consequence is small - I would say it is fairly uncommon to have a ';' in a folder name - there are no such in GECKO right now, and let's not add any.

johan-gson:
So, to explain how you access the functions in the adapter:
function model = applyKcatConstraints(model,kcatSource,kcatComplex,kcatSubs,updateRxns,modelAdapter)
%...
   params = modelAdapter.getSpontaneousRxns(model);
%...
end

Example of how to call this:
adapter = Yeast8Adapter();
x = applyKcatConstraints(..., adapter);

or just 
x = applyKcatConstraints(..., Yeast8Adapter());

If you want to change some params in the adapter (which is why I left the params public, which is a bit ugly, but practical):
adapter = Yeast8Adapter();
adapter.params.whatever = 3;
x = applyKcatConstraints(..., adapter);

GECKO here knows nothing about the specialized adapters, the modelAdapter sent in is just an object of a class that inherits the ModelAdapter base class. Let's look at Yeast8Adapter as an example - this class inherits modelAdapter. Since getSpontaneousRxns is abstract, YeastAdapter must implement that function - otherwise you will get an error when you try to create an object of that class. When the call is made, matlab will always use the function in the specialized class first, and GECKO doesn't have to know anything about that implementation in the specialized class - this could for example be placed in another repo. So, in java or c++, this is how you must solve it since those are typed languages. In matlab, you would technically not need a base class to accomplish this, as long as the object sent in has the right functions it will work anyway. But I think it is still better to use a base class - this makes it clear what needs to be implemented in the adapter and is what programmers would expect, and it also allows for default implementations in the base class (see getParameters for example).

johan-gson:
GeckoInstaller is persistent - it calls savepath, so if you restart matlab the path is still changed. The same could be added to RAVEN, but I think you have something there already perhaps? I always have these installer classes on my disk for Human-GEM, RAVEN and Yeast-GEM

johan-gson:
About where to place the specialized adapters: I'm thinking that during development of GECKO3 it is probably nice to have them in the GECKO repo, since we then don't need to change stuff in multiple repos and worry about having them synchronized. I do think it makes sense to then move them to another repo - my suggestion would be to place them in the GEM repos - so, the HumanGEMAdapter would go into the Human-GEM repo, etc. If someone makes a change, such as change how spontaneous reactions are stored in that repo, the adapter needs to be changed as well. One downside of that is that we cannot then (well, it would be weird) make YeastAdapter the default choice when calling functions, since GECKO will then not know about it. Technically it is still possible, but ...

edkerk:
> So, to explain how you access the functions in the adapter: function model = applyKcatConstraints(model,kcatSource,kcatComplex,kcatSubs,updateRxns,modelAdapter) %... params = modelAdapter.getSpontaneousRxns(model); %... end

Ok, what confused me is when you run `adapter = Yeast8Adapter();`, in MATLAB's Variables window `adapter` only contains `params`, but there is no (visual) indication that it also contains functions. But indeed, typing `adapter.` and clicking `Tab` also shows the function names (in Windows anyway).

GECKOInstaller.m
mihai-sysbio: `;` are allowed in file paths in some operating systems, so unlikely but still possible.
I was thinking of a simple check if the path [contains](https://www.mathworks.com/help/matlab/ref/contains.html) `;` then exit with error message that `;` are not supported in the file paths.

johan-gson: The problem is this: I call a matlab function which returns all subpaths in a directory as a string, where each subpath is separated by semicolon. There is no easy way to tell if a semicolon represents a separator or if that semicolon is part of a directory name. Let's not spend more time on this, it is only a theoretical problem. Remember that the purpose of this function is to install GECKO - and we don't have code in any folders containing semicolons, and it is very unlikely that we will add such, so this is not a real problem.

mihai-sysbio: The `;` can also appear one folder above the GECKO installation, so while not a GECKO-caused problem, it would still lead to a problem for the user. The `contains` could thus be applied to the `sourceDir` of the install and uninstall functions, and then I would feel confident in removing the comment that `;` in paths would lead to failures.

Also, looking at the [`genpath` documentation](https://www.mathworks.com/help/matlab/ref/genpath.html) and testing in R2021b, the separator and problematic character is in fact `:` not `;`. Perhaps it would better to define this as a PATH_SEPARATOR, since it is reused in several lines.

johan-gson: Ok, fair point, fixed.

GECKOInstaller.m
```diff
@@ -0,0 +1,49 @@
...
+		end
+
+        function newPaths = GetFilteredSubPaths(path_, filter_)
+			%Check that there are no semicolons in the path - that will cause 
+			%problems since that is the separator used to separate paths
+			if contains(path_, ';')
+				error('The path in which GECKO resides may not contain semicolons for this installation to work!');
+			end
+            paths = genpath(path_);
+            splitPaths = strsplit(paths, ';');
```
mihai-sysbio, line 36:
```suggestion
            splitPaths = strsplit(paths, GECKOInstaller.PATH_SEPARATOR);
```

GECKOInstaller.m
```diff
@@ -0,0 +1,49 @@
...
+            %remove the last, it is empty
+            splitPaths = splitPaths(1,1:end-1);
+            matches = regexp(splitPaths, filter_, 'match');
+            okPaths = cellfun(@isempty, matches);
+            pathsLeft = splitPaths(1,okPaths);
+            newPaths = strcat(char(join(pathsLeft,';')),';');
```
mihai-sysbio, line 42:
```suggestion
            newPaths = strcat(char(join(pathsLeft, GECKOInstaller.PATH_SEPARATOR)), GECKOInstaller.PATH_SEPARATOR);
```

GECKOInstaller.m
```diff
@@ -0,0 +1,49 @@
...
+        end
+    end
+    
+    properties (Constant)
+      FILE_FILTER = '.*\.git.*';
+   end
```
mihai-sysbio, line 48:
```suggestion
      FILE_FILTER = '.*\.git.*';
      PATH_SEPARATOR = ';';
   end
```

mihai-sysbio (commented):
@johan-gson I think the previous commit [a27b88c](https://github.com/SysBioChalmers/GECKO/pull/168/commits/a27b88c0fa7342a07ff09974bd01bf81da8ddb7d) was also meant to address that the separator should be `:` instead of `;`, so I provided specific suggestions for this. Or maybe I just got confused?

GECKOInstaller.m
mihai-sysbio: > Also, looking at the [`genpath` documentation](https://www.mathworks.com/help/matlab/ref/genpath.html) and testing in R2021b, the separator and problematic character is in fact `:` not `;`.

I'm not sure this was (meant to be) addressed? So I added some suggestions below. Therefore I'm resolving the comment thread here to continue the discussion below if necessary.

johan-gson:
Ok, I made a final change to GECKOInstaller - I asked matlab for the separator instead of hardcoding ';' - I actually think the previous code only worked on Windows...

mihai-sysbio:
> Ok, I made a final change to GECKOInstaller - I asked matlab for the separator instead of hardcoding ';' - I actually think the previous code only worked on Windows...

Great improvements @johan-gson! I am happy to approve the PR, but I think it would be useful to get a second pair of eyes to look at it before merging.

mihai-sysbio (approved):
Gr

Issue #156, opened 2022-05-25
makeEcModel: model-specific kcat data in dedicated structure


edkerk:
### Description of the new feature:
Instead of having k<sub>cat</sub> values only directly integrated as stoichiometric coefficients, keep a separate data structure with model-specific k<sub>cat</sub> values.

Each field in the `model.kcat` structure has the same length, where each entry corresponds to a unique k<sub>cat</sub> value, that can be assigned to a unique combination of reaction, metabolite and enzyme.

The purpose of this structure is that it can be populated from different sources, (e.g. DLKcat, BRENDA/SABIO-RK, manual curation) and changed at any time, while these values can then be "applied" to the ec-model to update the k<sub>cat</sub> values that are used.

Take particular notice:
1. If a reaction has multiple co-substrates, there _can_ be multiple entries, each with a different `.metNames`-associated k<sub>cat</sub> value [a later function might then choose the lowest, highest or mean value to use as k<sub>cat</sub> for that reaction in the ec-model].
 
2. If a reaction has multiple genes associated (whether via AND or OR relationships), they can all have a separate entry. Downstream functions will again choose which of the provided k<sub>cat</sub> values will be used to populate the model.

Field | Type | Description
-- | -- | --
`model.kcat.rxns` | string | Reaction   IDs gathered directly from model
`model.kcat.eccodes` | string | Gathered   directly from model, can be further populated via e.g. Uniprot/KEGG?
`model.kcat.metNames` | string | Gathered directly from model, each substrate separatedly. If reaction is in reverse (.reverse=true), then product is shown
`model.kcat.metSmiles` | string | Parsed,   DLKcat has code for this?
`model.kcat.gene` | string | Gathered   directly from model, consider gene association (and/or rules)?
`model.kcat.uniprot` | string | Can be   parsed via Uniprot dump with GECKO code? Perhaps not necessary
`model.kcat.mw` | numeric | Parsed   from Uniprot/KEGG? Otherwise, predict from protein FASTA?
`model.kcat.reverse` | logical | Gathered   directly from the model, reverse reactions will have double entries, of which   half of them have reverse=true
`model.kcat.kcat` | float | Gathered   from various sources, see later. One value for each rxns-substrate pair.
`model.kcat.source` | numeric | 0 =   manual; 1 = brenda (or similar); 2 = kmax; 3 = DLKcat
[...]

PR #170, opened 2022-07-08
feat: applyKcatConstraints and GECKO light

edkerk:

- feat:
  - `makeEcModel` only 1 kcat per reaction (not multiple entries for multiple substrates)
  - `applyKcatConstraints` can take the new structure as generated by `makeEcModel` (see also https://github.com/SysBioChalmers/GECKO/issues/156#issuecomment-1174441460 for the content of `model.ec`)
    - TODO: for GECKO light, the units of enzyme usage need to be corrected, and alternative isoenzymes need to be considered.


geckomat/change_model/applyKcatConstraints.m
```diff
@@ -80,90 +41,60 @@
 end
 
 %Clear existing incorporation of enzyme usage
+% TODO: modify for geckoLight
```
johan-gson, line 44: Doesn't the code below also solve this for GeckoLight? It is only the stoichiometry of one metabolite, the prot_pool, that needs to be cleared, and I think the code does that?

geckomat/change_model/applyKcatConstraints.m
```diff
@@ -80,90 +41,60 @@
...
+    %Replace rxns and enzymes with their location in model
+    [~,newKcats(:,1)] = ismember(model.ec.rxns(newKcats(:,1)),model.rxns);
+    [~,newKcats(:,2)] = ismember(strcat('prot_',model.ec.enzymes(newKcats(:,2))),model.mets);
+    linearIndices     = sub2ind(size(model.S),newKcats(:,2),newKcats(:,1));
+    model.S(linearIndices) = -newKcats(:,4); %Substrate = negative
+    
+else %GECKO light formulation, where prot_pool represents all usages
+    newKcats=zeros(numel(updateRxns),1);
+    for i=1:numel(updateRxns)
+        %TODO:The current code is incorrect, it takes the sum of all
```
johan-gson, line 84: So, we need to have all alternative kcats with us here (i.e., the kcat for each isozyme), and use the largest of those. So, the model.ec.kcat needs to support multiple numbers for Gecko light. These structures could be different for Gecko light, I don't think there is much common code between the two ways to lay out the model?

geckomat/change_model/applyKcatConstraints.m
```diff
@@ -1,65 +1,26 @@
...
-%               dlkcat  DLKcat derived kcat values are used
-%               kcatdb  kcat values derived from BRENDA and SABIO-RK
-%                       databases are used
-%               manual  manual curated kcat values are used
-%               If 'manual' should be preferred, followed by 'kcatdb' and
-%               'dlkcat', kcatSource can be specified as 
-%               'manual>kcatdb>dlkcat'. Variations of this are possible.
-%   kcatComplex which kcat should be used if multiple subunits are defined,
-%               can either be 'max' (default), 'min' or 'median'.
 %   updateRxns  if not all enzyme constraints should be updated, this can
```
johan-gson, line 11: Add in the description that the second param is optional

geckomat/change_model/applyKcatConstraints.m
```diff
@@ -80,90 +41,60 @@
...
-    switch kcatComplex
-        case 'max'
-            updateKcats(multipleKcatEntries) = max(updateKcats(multipleKcatEntries),[],'omitnan');
-        case 'min'
-            nonZeroKcat = updateKcats(multipleKcatEntries);
-            updateKcats(multipleKcatEntries) = min(nonZeroKcat(nonZeroKcat>0),[],'omitnan');
-        case 'median'
-            nonZeroKcat = updateKcats(multipleKcatEntries);
-            updateKcats(multipleKcatEntries) = median(nonZeroKcat(nonZeroKcat>0),'omitnan');
+    newKcats(kcatLast+1:end,:)=[];
```
johan-gson, line 66: Perhaps for GeckoLight you could assign "max" as in the code you removed here. I realize it doesn't fit perfectly with the current code, but if you solve it here that way there is no need for special treatment of these reactions (i.e., reactions with GPRs containing OR (that has isozymes))

geckomat/change_model/makeEcModel.m
```diff
@@ -111,54 +104,28 @@
...
-%8: Make ec-extension structure, one for each kcat value
-emptyCell    = cell(numel(model.rxns)*5,1); %Prepare plenty of space
-emptyVect    = zeros(numel(model.rxns)*5,1);
-ec.rxns      = emptyCell;
-ec.eccodes   = emptyCell; % Can also be parsed from model. Only needed for classic GECKO matching
-ec.rxnEnzMat = zeros(numel(model.rxns)*5,numel(model.genes)); % Non-zeros will indicate the number of subunits
-%ec.substrate = emptyCell; 
-%ec.reverse   = logical(emptyVect); % Whether reaction is in reverse. Might be unnecessary, can also be gathered from the reaction ID ('_REV')
+ec.rxnEnzMat = zeros(numel(rxnWithGene),numel(model.genes)); % Non-zeros will indicate the number of subunits
 ec.kcat      = emptyVect;
```
johan-gson, line 124: Something needs to be done here for GeckoLight - the question is where to apply the max(kcats from isozymes).

johan-gson (approved):
This is as I see it moving in the right direction, but there are some things that needs to be fixed for Gecko light. I think the main question is where to apply the trick for isozymes, i.e., we need to estimate the MW/kcat of a reaction my taking the minimum MW/kcat from the isozymes. So where will this mimimum be calculated?

edkerk:
@johan-gson That would done in a separate function that can be run after `makeEcModel` and before `applyKcatConstraints`: `selectKcatValue` (see also https://github.com/SysBioChalmers/GECKO/discussions/169).

geckomat/change_model/applyKcatConstraints.m
edkerk: I wasn't 100% certain how GeckoLight model would look like, so had kept this TODO to check later if it is needed. But you're right, the current code should handle it properly.

geckomat/change_model/applyKcatConstraints.m
edkerk: Various functions can gather alternative kcats, before `selectKcatValue` selects the appropriate value (for DLKcat this is the max value for different substrates, for GeckoLight this is the min MW/kcat. These selected values are then entered into `model.ec.kcat`, so no need to support multiple kcat values there.

geckomat/change_model/makeEcModel.m
edkerk: `makeEcModel` is only making the basic structure, it is not yet populated with kcat values.

geckomat/change_model/applyKcatConstraints.m
johan-gson: Ok, sounds good

PR #173, opened 2022-09-18
feat: prepare DLKcat input and read its output

edkerk:
- feat:
  - `writeDLKcatInput` prepares the TSV file that can directly be used as input for DLKcat.
  - `readDLKcatOutput` can read the TSV file of DLKcat output and construct a `kcatList` structure which contains all the prediction results
  - `selectKcatValue` takes `kcatList` and uses this to populate `model.ec.kcat`. If multiple kcat values match the same reaction (because of multiple subunits and/or multiple substrates in the same reaction), it will choose one value (by default the highest)

Other comments:
- With later functions, `kcatList` can also be generated from e.g. GotEnzymes output, a BRENDA/SABIO-RK database, manual curation etc.
- A default list of metabolite names is provided, which `writeDLKcatInput` will ignore when generating the input file, as these metabolites are too small or form a pair of currency metabolites (e.g. ATP + ADP), as was also done in the DLKcat paper. This list is probably not exhaustive, but could be expanded over time. Users can also provide alternative lists of metabolites.
- `readDLKcatOutput` can add `metSmiles` to the model, which were found by DLKcat. Note that RAVEN [PR #445](https://github.com/SysBioChalmers/RAVEN/pull/445) is required for further handling of `metSmiles` field.
- After `selectKcatValue`, one can use `applyKcatConstraints` to translate the `model.ec.kcat` values to actual constrains in the model (according to the general GECKO3 [scheme](https://github.com/SysBioChalmers/GECKO/discussions/169)).


feiranl:
It would be great if we add SMILEs information for the currency mets in the same file.

edkerk:
> It would be great if we add SMILEs information for the currency mets in the same file.

You mean in `DLKCatCurrencyMets.tsv`?

feiranl:
Yes. Since not every model use the name as in this file. Adding SMILES information would be easier for the mapping.

edkerk:
It is indeed preferred to have it work with as many models as possible, but most models do not have SMILES, so just adding SMILES information would not help too much.

The idea was that `DLKcatCurrencyMets` could act as a list that is expanded with more commonly used metabolite names (note that it already includes ATP and adenosine triphosphate). In documentation we can mention that if a user notices that a particular metabolite is not included, they can make a GitHub issue and we'll add it in the repo as well. In addition, the user can provide their own list of metabolites.

One thing that can at least be done is making the matching of metabolite names case insensitive and ignore any whitespaces and special characters. Beyond SMILES, it could be better to then match to other annotations (KEGG, ChEBI, etc.).

I suggest to have this improvement being addressed in a future PR. The current functions should have the basic functionality, we can then improve on it further.

edkerk:
It now matches without spaces, special characters and ignoring upper/lower case.

To test this, you can load yeast-GEM (or any other model), and run the following code to only predict 100 kcat values (otherwise it takes a while)

```matlab
model = makeEcModel('model');
DLKcatIDs = writeDLKcatInput(model);
DLKcatIDs.rxns(101:end) = [];
DLKcatIDs.genes(101:end) = [];
DLKcatIDs.substrates(101:end) = [];
```
Before running DLKcat, keep only the first 20 searches: `head -101 DLKcatInput.tsv > DLKcatInput2.tsv`

Then, run the `DLKcatInput2.tsv` through DLKcat, following the example explained in step 6 [here](https://github.com/SysBioChalmers/DLKcat/tree/master/DeeplearningApproach#readme). Make sure you copy the `output.tsv` or modify the next function to direct towards its location.

```matlab
[kcatList, model] = readDLKcatOutput(model,'output.tsv',DLKcatIDs);
[model, rxnIdx] = selectKcatValue(model,kcatList);
```

You can then inspect that the kcat values have appeared in `model.ec.kcat`. `applyKcatConstraints` can then apply these kcat values to the S-matrix, but that is not the topic of this PR. This PR is just focused on `writeDLKcatInput`, `readDLKcatOutput` and `selectKcatValue`, to show its basic functionality, which can be improved in further PRs.

feiranl (approved):
Looks nice!

le-yuan (approved):
It looks very nice, thanks for your effort!

PR #180, opened 2022-10-02
feat: Implemented GECKO Light functionality in GECKO 3.

johan-gson:
Also adapted the code to be able to handle Human-GEM in addition to Yeast-GEM. I have not tested the code with Yeast-GEM. The code is not well tested at this point.


feiranl (approved):
LGTM! I have tested both light and normal version for Human-GEM and Yeast-GEM.

feiranl:
I just have one concern. The unit of added protein pool/enzyme in gecko light and normal gecko should be consistent(umol/gDW or mmol/gDW), which will be much easier for further analysis functions such as analyze protein pool usage. Ed, I totally agree  that we need to do the scale up to avoid small fluxes. As I understand, GECKO2 is converting `mmol to gram` in enzyme usage rxns, which causes small fluxes. We can actually just change the protein pool from gram to mg, then the enzyme usage from protein pool conversion will be from `mmol to mg`, which actually scales up 3 magnitude, and all units in most of reactions are in mmol, except the protein pool rxn.

geckomat/change_model/makeEcModel.m
```diff
@@ -8,16 +8,18 @@
 %   model.
 %
 % Input:
-%   model       a model in RAVEN format
-%   geckoLight  true if a simplified GECKO light model should be generated.
-%               Optional, default is false.
+%   model        a model in RAVEN format
+%   modelAdapter A modelAdapter, an instance of a class inheriting the 
+%                ModelAdapter class, e.g., YeastAdapter.
+%   geckoLight   true if a simplified GECKO light model should be generated.
```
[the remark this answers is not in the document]
johan-gson: I think it will work automatically? Or do you mean just in the  comment?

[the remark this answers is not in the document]
johan-gson: I think this will work automatically, or do you mean just in the comment?

geckomat/change_model/applyKcatConstraints.m
```diff
@@ -79,22 +79,37 @@
...
+                nextIndex = nextIndex + 1;
+                if  nextIndex > length(model.ec.rxns) || ~strcmp(model.rxns(i), model.ec.rxns(nextIndex))
+                    break;
+                end
+            end
+            if (updated)
+                %Light model: always use the "cheapest" isozyme, that is what the 
+                %optimization will choose anyway unless isozymes are individually constrained.
+                %TODO:Check that the units are correct.
+                model.S(prot_pool_idx, i) = min(newMWKcats) * 3600; %per second -> per hour
```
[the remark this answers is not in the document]
johan-gson: Agreed, I can fix that

johan-gson:
> I just have one concern. The unit of added protein pool/enzyme in gecko light and normal gecko should be consistent(umol/gDW or mmol/gDW), which will be much easier for further analysis functions such as analyze protein pool usage. @edkerk , I totally agree that we need to do the scale up to avoid small fluxes. As I understand, GECKO2 is converting `mmol to gram` in enzyme usage rxns, which causes small fluxes. We can actually just change the unit of protein pool from gram to mg, then the enzyme usage from protein pool conversion will be from `mmol to mg`, which actually scales up 3 magnitude, and all units in most of reactions are in mmol, except the protein pool rxn.

Agreed, could someone help me fix that or tell me exactly what I should change to get it right?

geckomat/change_model/applyKcatConstraints.m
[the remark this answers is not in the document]
johan-gson: Hmm, I'm a bit confused here. So, the protein pool reaction is the same for both light and full, so if we turn this around, which I think makes sense, I had it in that order in the brain modeling for example, we also need to turn this around in the full model. So, there is a sign that needs to be changes somewhere there as well. Could any of you advise which place would be best, I have some problem understanding the consequences?

johan-gson:
Ok, seems I lost write access to the repo, I sent a message to Ivan, but any changes may be delayed a bit.

edkerk:
> Ok, seems I lost write access to the repo, I sent a message to Ivan, but any changes may be delayed a bit.

Should be fixed now.

johan-gson:
ok, thanks, it worked now!

geckomat/change_model/makeEcModel.m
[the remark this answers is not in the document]
feiranl: No, currently, `1 `does not work
Screen Shot 2022-11-12 at 9 53 30 AM

geckomat/change_model/applyKcatConstraints.m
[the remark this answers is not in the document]
feiranl: in the normal gecko of this function, enzyme is added as substrates, so here this line should be  `model.S(prot_pool_idx, i) = min(newMWKcats) * (-3600)`; @Yu-sysbio @edkerk either you may need to double check this.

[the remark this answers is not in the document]
Yu-sysbio: > 

It would be better to keep the same structure as GECKO1&2.

Here show reactions in ecYeast (created by GECKO1&2) as an example:
```
prot_pool_exchange	 -> prot_pool[c]
draw_prot_P15454	20.6371 prot_pool[c]  -> prot_P15454[c]
r_0330No1	s_0434[c] + s_0615[c] + 3.0864e-06 prot_P15454[c]  -> s_0394[c] + s_0613[c]
```
In GECKO1&2, `prot_pool` is as the product in **prot_pool_exchange** reaction and as the substrate in **draw_prot_xxxx** reaction. Then the product of the **draw_prot_xxxx** reaction is as the substrate in the corresponding metabolic reaction.

In GECKO3, currently it seems different, which I think should be corrected.
Regarding the light version, `prot_pool` is as the SUBSTRATE in **prot_pool_exchange** reaction and as the PRODUCT in the corresponding metabolic reaction, which is just the opposite (see reactions below).
```
prot_pool_exchange	prot_pool[c] -> 
r_0330	s_0434[c] + s_0615[c] -> s_0394[c] + s_0613[c] + (mw/kcat) prot_pool[c]
```

Regarding the full version, `prot_pool` is as the SUBSTRATE in **prot_pool_exchange** reaction and as the PRODUCT in **draw_prot_xxxx** reaction. However, the substrate in **draw_prot_xxxx** reaction is also the substrate in the corresponding metabolic reaction, which is incorrect (see reactions below).
```
prot_pool_exchange	prot_pool[c] -> 
draw_prot_P15454	20.6371 prot_P15454[c]  -> prot_pool[c]
r_0330No1	s_0434[c] + s_0615[c] + 3.0864e-06 prot_P15454[c]  -> s_0394[c] + s_0613[c]
```

Could consider my suggestions on the correction:
Regarding the light version, in **prot_pool_exchange** reaction change the stoich coeff of `prot_pool` to 1, and in metabolic reactions change `prot_pool` to negative.
Regarding the full version, in **prot_pool_exchange** reaction change the stoich coeff of `prot_pool` to 1, and in **draw_prot_xxxx** reaction change `prot_pool` to negative and `prot_xxxx` to 1 (note that MW should be as the stoich coeff of `prot_pool` rather than `prot_xxxx`, which should also be corrected).

geckomat/change_model/makeEcModel.m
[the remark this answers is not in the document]
johan-gson: When looking at the code, I actually think it should be as it is, that we enforce to use a boolean (logical). Ed explicitly made an effort here :). I know that people in MATLAB often use 1 or 0 (and also in c code), but I think that is a bad habit, true or false is in my mind more clear. But if people think it is a problem, we can remove it, this is not important to me. What do you think @edkerk, do you have some kind of general strategy for this (since you added the check)?

geckomat/change_model/applyKcatConstraints.m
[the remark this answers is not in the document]
johan-gson: @edkerk I think you did this change, right, from version 2 to 3? Did you have a good reason for this, or did it just "happen"? I'm just thinking it would be good to know before I go in and change it.

geckomat/change_model/applyKcatConstraints.m
```diff
@@ -79,22 +79,37 @@
...
-        %TODO:Correct units.
-    end
-    %All draw directly from protein pool
+    nextIndex = 1;
     prot_pool_idx = find(ismember(model.mets,'prot_pool'));
-
-    [~,rxnIds] = ismember(model.ec.rxns(updateRxns),model.rxns);
-    model.S(prot_pool_idx,rxnIds) = -newKcats; %Substrate = negative
+    for i = 1:numel(model.rxns)
+        if nextIndex <= length(model.ec.rxns) && strcmp(model.rxns(i), model.ec.rxns(nextIndex))
```
mihai-sysbio, line 85: I believe this could be done faster by sorting by `ec` and then `ec.rxn`

geckomat/change_model/makeEcModel.m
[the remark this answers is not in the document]
edkerk: `1` and `0` as `true` and `false` should indeed be allowed, will modify the check and consistently apply this whenever a logical input is expected.

geckomat/change_model/applyKcatConstraints.m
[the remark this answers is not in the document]
edkerk: The stoichiometric coefficients in the draw_prot and prot_pool reactions are indeed incorrect in the full version, this is a bug that hadn't been noticed yet, as the GECKO3-derived ec-models haven't yet been tested. Put this will indeed need to be fixed in the relevant functions.

edkerk:
> I just have one concern. The unit of added protein pool/enzyme in gecko light and normal gecko should be consistent(umol/gDW or mmol/gDW), which will be much easier for further analysis functions such as analyze protein pool usage. @edkerk , I totally agree that we need to do the scale up to avoid small fluxes. As I understand, GECKO2 is converting `mmol to gram` in enzyme usage rxns, which causes small fluxes. We can actually just change the unit of protein pool from gram to mg, then the enzyme usage from protein pool conversion will be from `mmol to mg`, which actually scales up 3 magnitude, and all units in most of reactions are in mmol, except the protein pool rxn.

I think this would indeed also work!

geckomat/model_adapter/ModelAdapter.m
```diff
@@ -17,7 +17,12 @@
         %end
         
 		function model = manualModifications(obj,model) %default is to do nothing
-		end
+        end
+        
+        %The genes returned here should match the gene id set in the parameter params.uniprotGeneIdField
+        function genes = getUniprotCompatibleGenes(obj,model)
+            genes = model.genes;
+        end
```
edkerk, line 25: This should also be in `YeastGEMAdapter.m`, right?

geckomat/model_adapter/HumanGEMAdapter.m
```diff
@@ -42,8 +52,33 @@
...
 		
+        %Ensembl gene ids are not available in uniprot (ensembl returns transcripts, not genes)
+        %Therefore, we collect gene symbols from uniprot, and need to convert the ensembl genes
+        %here to gene symbols as well
+        function genes = getUniprotCompatibleGenes(obj,model)
+            % get the path
+            tmpfile = fullfile(getHumanGEMRootPath(),'model','genes.tsv');
+
+            % import as structure, convert to table, and extract header
+            tmp = struct2table(importTsvFile(tmpfile));
```
edkerk, line 65: No major comment, as this only affects human-GEM. But these lines (`genes.tsv` file and `importTsvFile` function) require the full human-GEM repository to be present (and in MATLAB path). This should be clearly communicated before/when using this adapter, as users might just come with the human-GEM.xml file instead.

geckomat/utilities/loadDatabases.m
```diff
@@ -1,29 +1,37 @@
...
+%    fclose(fid);
+%    gunzip(fullfile(geckoPath,'databases/uniprot/',filename))
+%    clear uniprotDL % Load file fresh
 else
-    disp(['Load existing Uniprot database file for taxonomic ID ' taxonID '.'])
+    disp(['Loading existing Uniprot database file for taxonomic ID ' taxonID '.'])
 end
 
-fid         = fopen(fullfile(geckoPath,'databases','uniprotDB.tsv'),'r');
+fid         = fopen(fullfile(geckoPath,'databases/uniprot/',filename),'r');
```
edkerk, line 34: Not for this PR, but ideally any model-specific files should not be stored in the GECKO folder structure. The ModelAdapter could include a parameter that specifies the path where model-specific files should be stored? This also affect other functions (for instance `getComplexData`), and is probably best addressed in a separate PR.

geckomat/model_adapter/ModelAdapter.m
johan-gson: So, if I remember correctly, the yeast model follows the standard behavior, so there is no need to specialize any code for that and we can therefore rely on the base class implementation. I have been running them both and they worked, so I think it should be ok.

johan-gson: There was some kind of special case for Human-GEM that I had to deal with.

geckomat/model_adapter/HumanGEMAdapter.m
johan-gson: I have always assumed that the whole Human-GEM is available for everything I do. Sure, this needs to be communicated.

geckomat/model_adapter/ModelAdapter.m
edkerk: But it should be included, as `makeEcModel` is calling `uniprotCompatibleGenes = modelAdapter.getUniprotCompatibleGenes(model);`, and for simples cases this should just output `model.genes` as is done in this template adapter.

geckomat/utilities/loadDatabases.m
ae-tafur: I agree that model-specific files should be stored in an "external" folder. I suggest this structure. 

Screenshot 2022-11-30 at 10 01 13

the idea is that projects folder (can be also call Custom ? ) contains the data/scripts for model-specific files it a path is not defined in getModelParameters.m. 

getModelParameters.m also includes a path where to store the results ecModel/prot_model, this is usefull when working with git repos locally. 

I already implemented this approach by using GECKO 2 (in my fork). Maybe it can be implemented for GECKO 3 ?

johan-gson: So, in my mind, it is ok to have it there, but it could also be somewhere else as long as it is a place that we can guarantee will be available. Keep in mind that these files are not part of the repo - they are downloaded on the fly - it is only the folder that is available. The code checks for if the file with the same name as the taxon id is in the folder, cached there from previous runs. If not, it downloads and saves the file in the folder. So, this is in my mind not model-specific in that sense, it is only the cached files that are downloaded that are specific.

ae-tafur: But I mean, not only one specific file, GECKO also have many model/oganism specific functions which I think should be in other folder. I mean there should be a src folder with the GECKo functions that no need change by the user

edkerk: Let's have this discussion in a separate Issues, will open one now.

johan-gson:
So, I'm not sure I get what you mean fully, but will explain anyway :).
Since YeastGEMAdapter inherits ModelAdapter, it will automatically have all
the functions that are in ModelAdapter through the inheritance. What we do
in HumanGEMAdapter is to *override *the function
getUniprotCompatibleGenes(obj,model)
and provide a different implementation. If we had not done
that, HumanGEMAdapter and YeastGEMAdapter would both have the
implementation of the base class, which just returns the genes as is. Since
that strategy doesn't work for HumanGEMAdapter, we provide a different
implementation. This is (some of) the power of object-oriented programming,
that you can inherit something and override the things you don't like.

geckomat/utilities/loadDatabases.m
edkerk: Now in #187. This discussion is a bigger issue, but should not keep this PR from merging.

geckomat/change_model/applyKcatConstraints.m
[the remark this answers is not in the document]
edkerk: Fixed in 44e9ffd for full GECKO, and ce72bb1249f0245fc96f05608deb8a276a4eca83 for GECKO light.

edkerk:
@johan-gson I've made a separate branch (https://github.com/SysBioChalmers/GECKO/tree/feat/fuzzyMatching) with the fuzzy-matching commit and removed it from this PR, otherwise the current PR becomes gigantic. It would be good to finish this PR and merge it to `devel3`, and then continue from that.

edkerk:
In full GECKO, `model.ec.kcat` only contains kcat values that are actually incorporated in the model (and this is also reflected in `model.ec.rxns` etc.). `selectKcatValue` can take a `kcatList`, which can contain multiple kcat values per reaction (because of multiple substrates, or multiple isozymes & subunits), and from this `kcatList` select one kcat value per reaction to make it into `model.ec.kcat`. All that `applyKcatConstraint` does it take the value in `model.ec.kcat` and turn this into a coefficient in the S-matrix.

GECKO light currently has all isozymes as separate entries in `model.ec.kcat`, where `applyKcatConstraint` selects the highest kcat value across enzymes. For consistent usage, `selectKcatValue` should do this in GECKO light: take all predicted kcat values per reaction (across isozymes) and take the highest value across isozymes. Which isozyme this came from can be documented in `model.ec.notes`.

@johan-gson Do you have a strong reason why the current approach would be preferred? Otherwise it would be best to refactor and have consistent behaviour across the two model types. `applyKcatConstraints` could be called multiple times when repeatingly simulating the model with slightly different kcat values, so it would be beneficial to move this selection of highest-across-isozymes to `selectKcatValue` instead.

johan-gson:
The problem is that GECKO light doesn't take the largest kcat, it takes the smallest MW/kcat, so if the enzymes have different MW this may not be right. This is why I propose to either add an extra mapping field or do something with the rxn id for the light model.

edkerk:
Clear answer, then it's left like it is.

Issue #160, opened 2022-05-25
loadComplexData: read Complex Portal or similar input for subunit numbers


edkerk:
### Description of the new feature:
If the data is available, it could be taken into account that not all subunits are necessarily present in equal amounts in complexes. This will involve:
- Gathering the protein complex data (e.g. from [Complex Portal](https://www.ebi.ac.uk/complexportal))
- Define how to represent the complex data in a suitable way in the model structure (not trivial, as adding stoichiometric coefficients in `grRules` are not ideal as they'd need to be parsed out each time, while something like `rxnGeneMat` would not work if there are alternative configurations of a complex (e.g. `(2 subunit A + 1 subunit B) or (2 subunit A + 1 subunit C)`). **edit**: the `rxnGeneMat` option would work if isoenzymes are first split by `expandModel`
- Consider that kcat values that are reported are (typically) for the whole functional complex, not for individual subunits

edkerk:
The number of subunits is now included in the `model.ec` structure in #156 and is considered when applying the kcat constraints when using [`applyKcatConstraints`](https://github.com/SysBioChalmers/GECKO/blob/feat/makeEcModel/geckomat/change_model/applyKcatConstraints.m) from [feat/makeEcModel](https://github.com/SysBioChalmers/GECKO/tree/feat/makeEcModel) branch.

It does not require the `rxnGeneMat` strategy as suggested above.

At the moment it assumes a subunit stoichiometry of 1:1(:1:1... etc.). Gathering and parsing Complex Portal data is not yet implemented. This could first be accomplished for organisms that are present in Complex Portal. On top of that, one could imagine complex prediction based on sequence similarity, but that would be implemented much later.

edkerk:
`applyKcatConstraints` can deal with the subunit information in `model.ec.rxnEnzMat` since PR #170, so this Issue is rather about populating the `model.ec.rxnEnzMat` structure with subunit information from e.g. Complex Portal.

ae-tafur:
Obtaining the information from the complex portal whenever required may take some time. I already have a `getComplexData` function to download all the information from the database, or from a specific organism available in complex portal, and save it in a JSON file. Following this structure:

`[
  {
    "complexID": "CPX-3244",
    "name": "SCF-Das1 ubiquitin ligase complex",
    "specie": "Saccharomyces cerevisiae; 559292",
    "geneName": [
      "CDC34",
      "DAS1",
      "HRT1",
      "SKP1",
[...]

PR #182, opened 2022-10-21
feat: get and apply complex data from complex portal

ae-tafur:

Solution to #160

Add two functions:

1. getComplexData.m to get curated information for complex in complex portal, and save it in a .json file. 
2. applyComplexData.m which read the .json file with the complex data and apply it to populate the model.ec.rxnEnzMat field in the model. For now, I suggest to run applyComplexData after makeEcModel since it returns the proposed complex information for manual curation as suggested [here](https://github.com/SysBioChalmers/GECKO/issues/160#issuecomment-1267503120)


geckomat/change_model/getComplexData.m
```diff
@@ -0,0 +1,139 @@
...
+            % then assign an incorrect stochiometry 
+            if numel(idxIntType) > 1
+                stochiometry = split({temp.participants(idxIntType).stochiometry}.', ',');
+            else
+                stochiometry = split({temp.participants(idxIntType).stochiometry}.', ',').';
+            end 
+            values = str2double(erase(stochiometry(:,1),"minValue: ")).';
+            complexData(i,6) = {values};
+        else
+            complexData(i,6) = {repelem(0,numel(complexData{i,4}))};
```
edkerk, line 110:
```suggestion
            complexData(i,6) = {repelem(1,numel(complexData{i,4}))};
```
If no specific stochiometry is given, should it not be filled with 1s instead of 0s? For [CPX-3248](https://www.ebi.ac.uk/complexportal/complex/CPX-3248) no stochiometry is given, in contrast to e.g. [CPX-3238](https://www.ebi.ac.uk/complexportal/complex/CPX-3238). By setting 0s and running `applyComplexData` afterwards results in subunits no longer being required for those reactions. By setting 1s we just keep the status quo: we do not know the exact stochiometry so we assume 1:1.

edkerk (changes requested):
Looks really good (except for the one comment).

The location of the json file is fine in database for now, but should ideally use a location that is specified in the model adapter. But this can be implemented in a later PR foused on file locations.

geckomat/change_model/getComplexData.m
ae-tafur: Yes, I think about that, I keep to zero in order to notify the user about it. But maybe can be another extra column mentioning that status quo: so user can take this into account for potential manual curation

edkerk: Ah, I hadn't noticed you actually deal with this when complex data is applied to the model:

https://github.com/SysBioChalmers/GECKO/blob/90f1f9171b10930374a1b30f3b278cc9ba7a0036/geckomat/change_model/applyComplexData.m#L97-L101

But indeed, let's include a `stochiometryDefined` column (or similar) in the `foundComplex`, `proposedComplex` and `complexInfo` data structures, and in the `complex_data.json` file.

edkerk (approved):

edkerk (approved):
I made a few changes:
- getComplexData keeps track of whether stochiometry was defined in Complex Portal or not
- getComplexData expands complexes-of-complexes
- getComplexData can write complex_data.json to any path
- applyComplexData can take complexInfo directly (from getComplexData)
- documentation of the two functions

Issue #157, opened 2022-05-25
feat: generate JSON general kcat database


edkerk:
### Description of the new feature:
By repurposing some of the DLKcat [code](https://github.com/SysBioChalmers/DLKcat/tree/master/DeeplearningApproach/Code/preprocess), construct a [JSON file](https://github.com/SysBioChalmers/DLKcat/tree/master/DeeplearningApproach/Data/database) similar as in DLKcat. This JSON contains k<sub>cat</sub> values from BRENDA and SABIO-RK databases, and could be run every 6-12 months.

In contrast to the current DLKcat code, some filtering steps should be skipped, to (a) keep specific activities; (b) keep cases where no amino acid sequence could be assigned to. By doing so, this kcat database can be used in GECKO's fuzzy matching approach. The file thereby replaces the `max_Kcat` and `max_SA` files that GECKO [currently uses](https://github.com/SysBioChalmers/GECKO/tree/master/databases).

This JSON file should be loaded into MATLAB, to be used in GECKO fuzzy matching (which is refactored in a separate Issue).

However, if the purpose of this file is only used for GECKO's fuzzy matching, then we might as well stick with the existing `max_Kcat` and `max_SA` files? Let's keep this Issue on hold for now.

mihai-sysbio:
I'm not totally sure on the scope of the issue: is it about creating a local k<sub>cat</sub> file or about how that file is used by GECKO? Perhaps it would be easier to discuss verbally and write up conclusions.

edkerk:
Both about the actual file and how it is used. But it sounds like GotEnzymes will also be able to provide such a file (did not know this when opening this issue), so this issue will most likely be addressed when GECKO and GotEnzymes can work together.

edkerk:
Does not strictly have to be JSON, but it should be flat-text so that it is diff-able.

edkerk:
For reference, [this](https://dev-metatlas.csbi.chalmers.se/api/v2/gotenzymes/enzymes?filters[organism]=sce&pagination[page]=1&pagination[pageSize]=500) is what GotEnzymes output looks like (showing only first 28 lines here:

```
{"enzymes":[{"gene":"YAL012W","organism":"sce","domain":"E","ko":"K01758","reaction_id":"R04770","ec_number":"4.4.1.1;4.4.1.11","compound":"C00109","kcat_values":1.1646},
[...]

PR #188, opened 2022-12-19
feat: fuzzy kcat matching

edkerk:
Added the classical GECKO 2 behavior to GECKO 3:
* If the EC Codes field is missing in the model, the EC codes will be filled into the model.ec structure using the classical GECKO 2 method.
* It is now also possible to use the fuzzyKcatMatching function to get kcats from BRENDA etc. using the fuzzy matching from GECKO2.


geckomat/get_enzyme_data/loadBRENDAdata.m
```diff
@@ -1,22 +1,23 @@
 %%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%%
- function [KCATcell, SAcell] = loadBRENDAdata
-     KCAT_file      = '../../databases/max_KCAT.txt';
-     SA_file        = '../../databases/max_SA.txt';
-     MW_file        = '../../databases/max_MW.txt';
+ function [KCATcell, SAcell] = loadBRENDAdata(modelAdapter)
+     KCAT_file      = modelAdapter.getFilePath('max_KCAT.txt');
+     SA_file        = modelAdapter.getFilePath('max_SA.txt');
+     MW_file        = modelAdapter.getFilePath('max_MW.txt');
```
edkerk, line 5: These files are not model specifric, so should be kept in the `databases` folder?

ae-tafur: > These files are not model specific, so should be kept in the `databases` folder?

I agree, they are `databases`

edkerk: Ah, I see that they are model specific, because they only include human kcat values. If that is necessary, it would be better to provide a function that can load the full generic database files and subset those for the desired organism, instead of providing more model-specific files.

ae-tafur: But this databases are also used to fill those reactions with the proteins information from other organism in case it is not reported for that organism, or if the model is over constrained

edkerk: Indeed, it does not replace the generic database files.

edkerk:
1. I copied commit 4866112 as part of PR [#189](https://github.com/SysBioChalmers/GECKO/pull/189), to separate the two issues (that commit can already be reviewed, while the fuzzy matching might need more refactoring & discussion). Once [#189](https://github.com/SysBioChalmers/GECKO/pull/189) is accepted it can also be merged into this PR, but at least [#189](https://github.com/SysBioChalmers/GECKO/pull/189) will not be unneccessarily delayed.

2. We should avoid to have too much model-specific files and duplication. See [above comment](https://github.com/SysBioChalmers/GECKO/pull/188#discussion_r1053179879), instead give `loadDatabase` a flag to only keep a selection of the loaded data.

3. Related to the above, `phylDist.mat` is duplicate of the file in `databases`.

4. The `protDatabase.mat` is partially superseded by the Uniprot text files that are obtained from `loadDatabases`. I'll refactor something similar for KEGG, so that we can avoid storing `*.mat` files in the repo (changes in these binary files cannot be tracked and they can quickly increase the size of the repo).

geckomat/get_enzyme_data/loadBRENDAdata.m
Yu-sysbio: Shall we still use the files `max_KCAT.txt`, `max_SA.txt` and `max_MW.txt` generated 4 years ago? And Brenda has already changed the manner to access the data, see [this](https://www.brenda-enzymes.org/download.php). The python part in GECKO1&2 that retrieves enzyme information from Brenda is obsolete.

edkerk: We can repurpose the DLKcat code that gathers this data?

feiranl: This kind of update of BRENDA is after DLKcat. I have code for gather and preprocess this data , but it is just functional not so clean.

edkerk: We would just run this locally every now and then (once a year?) and update the files in the repo, similar as KEGG-trained HMMs in RAVEN. So it's fine if the pipeline is not so clean at the moment, as long as the output is consistent. We can then write a function that can take the output (probably json?) and make the `max_*.txt` files.

edkerk:
Set this as draft as I'm currently refactoring the work, including the Uniprot and KEGG database files.

I'm not working on updating the `max_*.txt` files, which should potentially be done using DLKcat code. This is probably material for a separate PR.

Yu-sysbio:
> Not addressed is the actual fuzzy matching, or updated `max_*.txt` files. The latter could require some refactoring of the code, but this will be in a separate PR.

Regarding the updated `max_*.txt` files, I already have codes in my private repo, which can extract enzyme information from the latest Brenda database. The output files are however different from what GECKO1&2 needs, but I can modify the files afterward. 
It would be better to clarify in the separate PR: 1) do we need all three `max_*.txt` files, i.e., `max_KCAT.txt`, `max_SA.txt` and `max_MW.txt` in GECKO3? 2) should we keep the updated `max_*.txt` files in GECKO3 exactly the same structure as in GECKO1&2?

edkerk:
> Regarding the updated `max_*.txt` files, I already have codes in my private repo, which can extract enzyme information from the latest Brenda database. The output files are however different from what GECKO1&2 needs, but I can modify the files afterward. It would be better to clarify in the separate PR: 1) do we need all three `max_*.txt` files, i.e., `max_KCAT.txt`, `max_SA.txt` and `max_MW.txt` in GECKO3? 2) should we keep the updated `max_*.txt` files in GECKO3 exactly the same structure as in GECKO1&2?

Points 1) and 2) can be discussed in #157.

edkerk:
**Ready for review!**

This can now generate a BRENDA fuzzy matching based ec-model. Some code is available in `tutorials/protocol.m`. I have not checked whether the assigned kcat values are correct (should be reporting same as in GECKO1&2, as the matching approach hasn't changed). But the model can at least run FBA (without protein pool constraint, so same growth rate is reached as non-ec-model).

To do in next PR:
- Write a function that can choose which of the fuzzy matched kcat values should be used in the model (use `kcatList.origin` and `kcatList.wildcardLvl`).
- Correct documentation for each function.

geckomat/get_enzyme_data/loadDatabases.m
```diff
@@ -0,0 +1,148 @@
...
+% Output:
+%   databases       contains .uniprot and .kegg structures, dependent on
+%                   which databases were selected.
+%
+% Usage: databases = loadDatabases(modelAdapter,selectDatabase)
+
+if nargin<2
+    selectDatabase = 'both';
+end
+geckoPath = findGECKOroot();
```
johan-gson: I realize this code is just copied from my previous code, but I think it would be better to move the retrieval of this path to the base model, so it can be overridden. This way, we could create a small test taxon placed in the test folder when we do test cases, and a TestModelAdapter would then point to that folder instead of the standard one. That also makes it possible to provide a custom such database if desired.

geckomat/get_enzyme_data/loadDatabases.m
```diff
@@ -0,0 +1,148 @@
...
+param=modelAdapter.getParameters();
+keggID=param.keggID;
+taxonID=num2str(param.taxonID);
+
+weboptions('Timeout',10);
+
+%% Uniprot
+if any(strcmp(selectDatabase,{'uniprot','both'}))
+    filePath = fullfile(geckoPath,'databases','uniprot',[num2str(taxonID) '.tsv']);
```
johan-gson: I'm thinking then that this would be replaced by something like

filePath = fullfile(modelAdapter.getUniprotPath(),[num2str(taxonID) '.tsv']);

johan-gson: And in the base adapter for example  (not tested):

        function path = getUniprotPath(obj)
            geckoPath = findGECKOroot();
            path = fullfile(geckoPath,'databases','uniprot');
        end

Someone should think through if there should be just one function for all paths like I did now (yes I was lazy :) ) or one function per path, I don't have a strong opinion here.

johan-gson: This way, that function could be overridden in a TestModelAdapter

geckomat/get_enzyme_data/loadDatabases.m
```diff
@@ -0,0 +1,148 @@
...
+    databases.uniprot.ID      = fileContent{1};
+    databases.uniprot.genes   = fileContent{2};
+    databases.uniprot.eccodes = fileContent{4};
+    databases.uniprot.MW      = str2double(fileContent{7});
+    databases.uniprot.seq     = fileContent{8};
+end
+
+%% KEGG
+if any(strcmp(selectDatabase,{'kegg','both'}))
+    filePath = fullfile(geckoPath,'databases','kegg',[keggID '.tsv']);
```
johan-gson: Same thing here.

johan-gson:
Good changes to this code Ed! I had a comment about moving some things to the adapter - we can either do that now or later. Since we start with the old GECKO code and gradually improve it, we cannot expect to fix all things at once, everything that is an improvement passes as I see it! But in the final perfect product, we should have thought through how to be able to test everything.

johan-gson (approved):
So, I accidentally left my comments outside of the review, but see them as optional.

Yu-sysbio (approved):
Well done Ed! Looks good in general. I ran the script `tutorials/protocol.m` smoothly. Besides that, I set ub for the total protein pool reaction at 0.1 and got an infeasible solution. This is due to extremely low kcat values in the model as when increasing from 0.1 to 10 I got 1e-4 growth rate. This means that we indeed need tuning in the next stage.

I actually got some errors or found some bugs when making ecGEMs for other organisms e.g., E. coli. These might be not so relevant here and we could discuss them later maybe in test/debug processes?

edkerk:
@Yu-sysbio Yes, I'd prefer to merge this PR, so that further work can built on this code already, while separately making smaller bug-fixes.

@johan-gson Excellent suggestion, will implement this in later PR. Just one thing, you seemed to have commented on an earlier commit, not on the final code: https://github.com/SysBioChalmers/GECKO/blob/b2f72b148291d971fe056c82f59067991522e606/src/geckomat/get_enzyme_data/loadDatabases.m#L30-L38
Your argument still holds though!

Issue #147, opened 2021-03-10
bug: very low fluxes, close to solver tolerance


edkerk:
### Description of the bug:
The units by which enzymes are incorporated (mmol and g/gDCW) results in low stoichiometric coefficients, low UB for protein exchange reactions constrained by proteomics measurements, and low fluxes through these exchange reactions. These values can get close to or exceed the tolerance that is set for the LP solvers (`1e-9`).

This seems not to be an issue for ecYeast, but can particularly play a role when high-sensitivity proteomics data is integrated, as it is able to measure very low abundances (sub nanomolar). Also large kcat values can result in very low fluxes. However, I've only encountered issues in proteome constrained models (but at the same time I don't do much analysis with the `_batch` model).

In selected models, running flux minimization (`solveLP(model,1)`) throws an error that no solution is found. I don't know enough of the solver to pin down what the exact problem is. What more surprised me were errors in `flexibilizeProteins`, where **increasing the UB of a measured protein exchange reaction by 10-fold, without changing anything else, resulted in a model that could no longer be solved** (again, not enough knowledge of the solver to pin down the exact problem). Particularly this last problem is worrying, as a solution was found with the original UB (although resulting growth rate was too low), but increasing UB 'breaks' the model.

#### Expected behavior:
It is better not to have fluxes close to the solver's tolerance.

Inherently I do not see any reason why the enzyme usage coefficients and UBs could not be multiplied by 1000. This way, the enzyme usage and measured protein exchange reactions are in micromole/gDCW/h (instead of millimole), while the protein pool exchange reaction is in mg/gDW (instead of g/gDW). 

#### Suggested solution:
A quick work-around to solve this for proteomics-integrated models is by making the following changes in `ecModel` and `ecModel_batch` before running `generate_protModels`, to change the usage coefficients 1000-fold:

```matlab
protIdx = find(startsWith(ecModel.mets,'prot_'));
pExcIdx = find(contains(ecModel.rxns,'prot_'));
for i=1:length(protIdx)
    enzRxnIdx = find(ecModel.S(protIdx(i),:)); % All reactions involving protein
    enzRxnIdx = setdiff(enzRxnIdx,pExcIdx); % Leave out protein exchange
    coeffs = ecModel.S(protIdx(i),enzRxnIdx)*1000;
    ecModel.S(protIdx(i),enzRxnIdx) = coeffs;
end
```

And make the following two changes in `generate_protModels.m` to also increase the UBs 1000-fold:
[...]

PR #189, opened 2022-12-20
fix: use MW/kcat instead of 1/kcat

edkerk:
As suggested by @Yu-sysbio, to resolve low fluxes (#147) and have a model formulation that is coherent across full and light GECKO versions, the enzyme usage coefficients are defined as MW/kcat, where MW is given in Dalton (and not kDa).

### DIFF
--- a/geckomat/change_model/applyKcatConstraints.m
+++ b/geckomat/change_model/applyKcatConstraints.m
@@ -49,30 +49,29 @@ if ~model.ec.geckoLight
 end %no clearing needed for GECKO Light, will be overwritten below
 %For normal GECKO formulation (full model), where each enzyme is explicitly considered
 if ~model.ec.geckoLight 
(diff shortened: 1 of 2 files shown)


ae-tafur:
But this should be also implemented in modifyKcats and manualModifications functions ?

Yu-sysbio (approved):
Looks good!

edkerk:
`manualModifications` will become obsolete as it is now, instead we can directly change the kcat values in `model.ec.kcat`, before applying them with `applyKcatConstraints`. One can have a separate kcatList with curated kcat values that should be applied before/after gathering the DLKcat values. Possibly this should have a function that can read kcatList from txt file?

`modifyKcats` indeed needs to be refactored for these changes, but that can be done in a separate PR focused on modifying kcats to reach realistic growth rates (there are different strategies: `modifyKcats`, but also the Bayesian approach from DLKcat, and probably an even simpler approach (`modifyKcats` looks in BRENDA for alternative kcats, but something similar should also work on DLKcat-values only).

Issue #187, opened 2022-12-17
store model-specific files away from GECKO code


edkerk:
### Description of the new feature:
In GECKO1&2, model-specific files were kept in `databases` and `models` subfolders. I propose that a better approach would be to keep this separate.

1. The GECKO folder should have the GECKO code.
2. The ModelAdapters are where custom parameters and functions are kept.
3. The user can choose themselves where they want to have model-specific files written (and read from).
4. The path to the model-specific files location can be specified in the ModelAdapter.
5. This affects:
    - uniprot database files
    - complex portal database files
    - ec-model files
    - potential log files with statistics around the construction process
    - ... likely other files
6. GECKO can still come with one example, for _S. cerevisiae_ (as was GECKO1&2), but this would be in an `example` subfolder of GECKO, and is not the location where files are written/read by default.

ae-tafur:
Following the conversation.

All this functions need are organism/model specific:
- changeMedia_batch.m
- getModelParameters.m
- manualModifications.m
- removeIncorrectPathways.m
- scaleBioMass.m
- sumBioMass.m
- sumProtein.m

This is organism specific data:

- abs_proteomics.txt
- chemostatData.tsv
- fermentationData.txt
- manual_data.txt
- prot_abundance.txt
- uniprot.tab
 
So the user need to change this functions in GECKO depending on the project or organism he is working on. 

I understand that modelAdapter switch the functions depending what the user needs, but this is and implementation for those what GECKO already have, but what if the user it is working in a model which is not in the adapter function ?

johan-gson:
I in general don't have much opinions when it comes to storage of files. One thing that I think we should keep in mind though is the existence of general solutions, such as the one where the phylogenic id is automatically downloaded and stored in a folder. An alternative would be to automatically download it and store it in a folder specified by the model adapter. I don't have a strong opinion here, although I think the current solution is ok.

johan-gson:
I leave this for you @edkerk and @ae-tafur to decide. I'm happy as long as it works and as long as there is no extra specification needed to be done by the user - the goal is that things should be as automatic and simple as possible for a standard run.

edkerk:
For completion, here is the suggestion by @ae-tafur as written in #180:

> I agree that model-specific files should be stored in an "external" folder. I suggest this structure. 
> 
[...]

PR #190, opened 2022-12-22
refactor: change file structure to store model-specific outside of GECKO code

ae-tafur:

This a solution for #187

- a new file structure is introduce, GECKO functions which no requiere of user modifications are now in `src` folder. Functions that requires modifications by the user are stored in `userData` folder. `userData` is not added Matlab path during GECKO installation, because the model-specific functions have the same name although they are different models (e.g. there is a `sumBiomass`  function for yeast and other `sumBiomass`  function for  _Pseudomonas_ )
- Optional, in `getModelParameters` can also be defined a where GECKO will find the organism/model-specific functions (parameters.userDataPath) and where to store the output (parameters.outputPath).
- `parameters = getModelParameters` frequently call by almost all the functions, is defined as a function input.
- To keep the structure, few changes were introduced to saveEcModel:

    1. `saveEcModel` is now part of `geckomat` code.
    2. fix to allow store model in RAVEN format in the same folder.
    3. use condition as model name during proteomics integration to avoid overwrite the prot_model. One folder by condition containing the model and other output is created automatically.

- If a model is save in RAVEN format, the reactions id (model.rxns) will be sorted by RAVEN `exportForGit`. Then, this introduce a bug in proteomics integration because ecModel_batch has an extra reaction (prot_pool_exchange), resulting in reactions that do not have the same index.


edkerk (approved):
Looks good regarding the change in file structure! Regarding parameters and model I/O, this might change a bit in separate PRs, but for now this is also fine.

PR #191, opened 2022-12-22
feat: global ModelAdapter

edkerk:
To provide model-specific parameters and functions, the user should modify the dedicated `ModelAdapter`.
- `ModelAdapter` is located in the `userData/$model` folder of interest.
- `setGECKOModelAdapter('pathToModelAdapterFile')` should be run to set the `ModelAdapter` as global variable: it will be available for all functions. Example: `setGECKOModelAdapter('userData/ecYeastGEM/ModelAdapter.m')`
- Each function that requires information from `ModelAdapter` should start with:

```matlab
global GECKOModelAdapter
params=checkGECKOModelAdapter(GECKOModelAdapter);
```

### Additional changes in this PR:
- Renamed `userData/$model/scripts` to more generic `userData/$model/code`
- Moved model-specific `ModelAdapter` files into model-specific `userData/$model/` folder. Additional benefit is that they can be name identical (when running `setGECKOModelAdapter`, one should provide the full path (or relative path inside the GECKO folder), to avoid ambiguity). `userData/` is not in MATLAB path, so duplicate file names is no problem.
- Provide example model-specific file structure at `userData/yourModel`.
- Some minor adjustments of various function parameters that somehow involved ModelAdapter.
- Renamed a few more files that are obsolete. These are GECKO1&2 functions that are no longer necessary and will be deleted before GECKO3 release. They will not be refactored (although part of their code might be used in other functions).

Left to do in separate PR:
- Only functions that are already refactored for GECKO3 were modified for global ModelAdapter, needs to be properly implemented for other functiosn once being refactored.
- `userData/yourModel/ModelAdapter.m` is currently the same as the one for yeast-GEM. Should probably be modified a bit, to include useful examples code and parameters, without necessarily being duplate of the yeast-GEM file.


ae-tafur (commented):
This is a great solution !

- Also the user can define in modelAdapter the path to any place. But in any case, the user should defined (no matters where, outside of GECKO folder or in `userData`) four folders (code, data, model, output ), right ? 

- Why not also mark as obsolete `sumProtein` and  `changeMedia_yeast` in `utilities`  in this PR ?

ae-tafur (approved):

edkerk:
> This is a great solution !
> 
> * Also the user can define in modelAdapter the path to any place. But in any case, the user should defined (no matters where, outside of GECKO folder or in `userData`) four folders (code, data, model, output ), right ?
At the moment it assumes that the four folders are in the path that the user defines. We could of course also allow the user to define custom separate folders. Will implement in separate PR.

> * Why not also mark as obsolete `sumProtein` and  `changeMedia_yeast` in `utilities`  in this PR ?
I/we haven't refactored the functionality around these functions yet. I've only marked obsolete those functions whose functionality we've certainly replaced already.

johan-gson:
Hi Ed,

I'm sorry to say that I really dislike this solution, it goes against everything they teach in computer science :). It is considered bad practice to use global variables, for good reasons. In my mind, the modelAdapter should be a parameter to the functions - I understand you think it makes it cumbersome to call the functions - but I think that is how it should be. The problems in general with global variables is that a function with the same parameters sent in may behave differently depending on some external state. 

For example, let's consider when you want to run a test case on a small test model. You would then create some kind of TestModelAdapter for that small model, where you adapt the parameters to the test. With the solution you propose here, you will need to change that global parameter when you run the test, and then change it back afterwards to restore the previous state. And then, sometimes, the test fails in the middle, and you will have messed up the global variable, and just don't notice. A similar problem was with the path in previous GECKO versions, if the code was interrupted you almost always had to do some cd tricks before you could call it again.

If you want to get rid of the parameter to the functions, an alternative could be to have a function called getDefaultModelAdapter, that returns the global variable. Then you make the modelAdapter parameter optional, and if it is missing (nargin < X), you assign it to getDefaultModelAdapter(). This is an ok solution as I see it, I very much suggest changing to that approach. This way, the code is not really dependent of that global variable, and there is never any need to change it for test cases etc. An alternative would be to have a class with a static (persistent) variable. An example from MATLAB docs, although I would use separate set/get: 
classdef StoreData
   methods (Static)
      function out = setgetVar(data)
         persistent Var;
         if nargin
            Var = data;
         end
         out = Var;
      end
   end
end

johan-gson:
@edkerk  see above

ae-tafur:
> For example, let's consider when you want to run a test case on a small test model. You would then create some kind of TestModelAdapter for that small model, where you adapt the parameters to the test. With the solution you propose here, you will need to change that global parameter when you run the test, and then change it back afterwards to restore the previous state. And then, sometimes, the test fails in the middle, and you will have messed up the global variable, and just don't notice. A similar problem was with the path in previous GECKO versions, if the code was interrupted you almost always had to do some cd tricks before you could call it again.

But you can clear the global variable before call it again.

johan-gson:
You could, but again, this is just not the ways to do things. Google on why you shouldn't use global variables and you will understand. In my example above: When you are using GECKO with for example Yeast8, you will set that global variable to YeastAdapter. Then you run a test case that fails halfways through. Then the global variable will have changed to TestModelAdapter without you realizing it. Then you run some other code, and will by mistake use the wrong adapter. But I think what I proposed above with a default parameter solves the issue, the user will not have to send in the adapter? The function could even warn if the variable is not set.

johan-gson:
Also, if you have code working with different models, for example a project when you compare the metabolism of animals or something, this becomes impractical. In that case, you would in the code I proposed choose to send in the right adapter in each call, instead of switching adapter in the global variable.

ae-tafur:
Then, why not just pass the parameters structure to the functions, as generated by getModelParameters?

johan-gson:
So, in a way that is part of what we do with the adapter. The difference is that with the adapter you can also supply code to do things, such as figuring out where things are stored locally on your disk, conversion of gene symbols in the model, etc. 

I just saw what you did with the adapters - to give all adapters the same name. This is not at all what I had in mind - I think we have different views of how this should work, and I didn't realize that structuring how the data was stored also involved this kind of changes. This is similar to GECKO2, that we replace code depending on which model we want to run GECKO for, and is not at all needed with the adapter approach. It becomes problematic if you want to work with multiple models for example. It would probably be good if the three of us had a short meeting about this so we know where we are heading. Sorry for not being more involved, I have had a lot of things to do, but I'm now back in Sweden. I'll contact you  guys on Teams!

edkerk:
I think we're all on the same page on how custom functions & parameters work, it's just about the exact implementation of how the ModelAdapter is called. We'll discuss this offline and propose the necessary changes in a new PR.

Issue #186, opened 2022-11-11
Detect conserved words in GECKO3 and gives a warning when they are occupied in the basic GEM


feiranl:
#### Expected behavior:
Mets in the basic model starting with conserved words such as 'prot_' should be detected at the beginning of makeEcModel and gives a warning.

**I hereby confirm that I have:**
- [ ] Tested my code with [all requirements](https://github.com/SysBioChalmers/GECKO) for running GECKO
- [ ] Done this analysis in the `master` branch of the repository
- [ ] Checked that a similar issue does not exist already

PR #192, opened 2023-01-10
fix: some small fixes on gecko3 functions

edkerk:
- fix:
  - `getComplexData` critical bug
  - `getECfromDatabase` and `getECfromGEM` if ec-codes are assigned to only a subset of the reactions, and no `model.ec.eccodes` field yet exists
  - `fuzzyKcatMatching` confirms that `model.ec.eccodes` exists
  - `fuzzyKcatMatching` does not have kcatInfo as optional output, as this is already in `kcatList.wildcardLvl` and `kcatList.origin`
  - `fuzzyKcatMatching` explanation of its input and output
  - `makeEcModel` error if model.mets contains `prot_` and/or model.rxns contains `draw_` (solves #186)
 

src/geckomat/change_model/makeEcModel.m
```diff
@@ -103,6 +103,16 @@
     ec.geckoLight=false;
 end
 
+%Check for conflicting reaction and metabolite identifiers
+conflictId = startsWith(model.mets,'prot_');
```
ae-tafur, line 107: This is validation of the traditional GEM, right ? Because ecModel uses `prot_` or `draw_`

ae-tafur (approved):

src/geckomat/change_model/makeEcModel.m
edkerk: Exactly, `model` at this point is the traditional GEM that is used as input. Already having metabolites and reactions with these suffixes would undoubtly cause problems later on (which might not be so easy to trace back), so best to avoid them (and it probably doesn't affect many models).

PR #193, opened 2023-01-14
feat/fix: Updated the adapter pattern and added a manual test file fo…

johan-gson:
Updated the way we handle the adapters. All functions now take in an optional adapter, and it is not sent in, a default adapter is used. The default adapter can be set (and then the adapter no longer needs to be passed around).


src/geckomat/model_adapter/ModelAdapterManager.m
```diff
@@ -0,0 +1,68 @@
...
+        %not really needed, we could access it directly. Not very nice, but practical
+        function adapter = getAdapterFromPath(adapterPath, addToMatlabPath)
+            if nargin < 2
+                addToMatlabPath = true;
+            end
+            
+            potentialAdapterFiles = dir(fullfile(adapterPath,'*.m')); %gets all wav files in struct
+            %Check that there is only one .m file
+            if length(potentialAdapterFiles) ~= 1
+                error('getAdapterFromPath: One, and only one .m file is expected in the folder. This file is expected to contain a ModelAdapter.');
```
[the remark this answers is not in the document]
edkerk: I retract this suggestion, implementing this for yeast-GEM made it clear to me that it is very convenient to have the model adapter file directly in the model root.

edkerk (approved):
Works great, tested both with human-GEM and yeast-GEM, both with and without setting default adapters. `tutorials/protocol.m` is also updated.

johan-gson:
Would be a good idea to put it in a subfolder called adapter, that only contains one .m file - that would be very clean - we could then only add that folder to the path.

johan-gson:
i.e., as you suggested

src/geckomat/model_adapter/ModelAdapterManager.m
```diff
@@ -0,0 +1,71 @@
+%Abstract Base class for adapters for different species
+classdef ModelAdapterManager 
+	methods(Static)
+        %not really needed, we could access it directly. Not very nice, but practical
+        function adapter = getAdapterFromPath(adapterPath, addToMatlabPath)
+            if nargin < 2
+                addToMatlabPath = true;
+            end
+            
+            potentialAdapterFiles = dir(fullfile(adapterPath,'*.m')); %gets all wav files in struct
```
ae-tafur, line 10: why not be more specific ? e.g. all the adapters should be named as *GEMAdapter.m, so if an user have more in that folder, that files are not added. We then in the repo instructions we mention that.

ae-tafur:
> Would be a good idea to put it in a subfolder called adapter, that only contains one .m file - that would be very clean - we could then only add that folder to the path.

I think is better to keep specific functions/class to be part of the same folder individually, also can be easier with keep/store the repo with other potential ecModels

johan-gson:
I'm open for such changes as long as they are robust and works, which I think all of these suggestions are.

src/geckomat/model_adapter/ModelAdapterManager.m
edkerk: I think that was a criticism @johan-gson had with the earlier implementation, that using the same name actually causes problems?

It is fine if the adapter is set as default and not specified as parameter when running a function, as when you set the default parameter you explicitly refer to a certain path. But if the adapter is specified as function parameter, it should be findable from the MATLAB path, and in that case it should be a unique name. Otherwise having multiple model-specific folders in MATLAB path would cause conflicts.

edkerk:
> Would be a good idea to put it in a subfolder called adapter, that only contains one .m file - that would be very clean - we could then only add that folder to the path.

But with the current implementation it already only adds the folder containing the ModelAdapter*.m file to MATLAB path, not any other sub-folders.

@ae-tafur Do I understand correctly that you suggest the implementation as it is done here? We can then merge this PR.

ae-tafur:
> > Would be a good idea to put it in a subfolder called adapter, that only contains one .m file - that would be very clean - we could then only add that folder to the path.
> 
> But with the current implementation it already only adds the folder containing the ModelAdapter*.m file to MATLAB path, not any other sub-folders.
> 
> @ae-tafur Do I understand correctly that you suggest the implementation as it is done here? We can then merge this PR.

Yes, I do

PR #194, opened 2023-01-14
feat: skip making exchange rxns reversible

edkerk:
1. When making a new ec-model, currently all reversible reactions are made irreversible.
2. The reason for making irreversible is to avoid negative enzyme usage (or in other words: enzyme pseudometabolites being produced instead of consumed).
3. Only reactions that will be extended with enzyme constraints need to be made irreversible, other reactions (e.g. exchange) do not.
4. Making reactions unnecessarily irreversible results in a larger model with numerous loops.

Therefore, the code here only makes reactions reversible if they are associated to a gene.

Note, #193 should be merged to `devel3` first, after which this PR should be based on `devel3`.

### DIFF
--- a/src/geckomat/change_model/makeEcModel.m
+++ b/src/geckomat/change_model/makeEcModel.m
@@ -152,7 +152,8 @@ end
 
 %4: Make irreversible model (appends _REV to reaction IDs to indicate reverse
 %reactions)
-model=convertToIrrev(model);
+rxnsWithGenes = sum(model.rxnGeneMat,2);
+model=convertToIrrev(model, model.rxns(logical(rxnsWithGenes)));
 
 %5: Expand model, to separate isoenzymes (appends _EXP_* to reaction IDs to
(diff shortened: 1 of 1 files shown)


johan-gson (commented):
I wonder if this is really a good idea? I understand the size gain, but this hinders us from for example adding a standard cost to all non-spontaneous, non-exchange reactions without GPRs? And, are you sure we are not counting on that the model is irrev somewhere? I'm not sure, just concerned :)

edkerk:
That's a fair point. So maybe only exchange reactions should be left out? Because exchange reactions are never enzyme-catalyzed (their transport reactions might).

johan-gson:
Reasonable, agreed

feiranl:
I also agree only exchange rxns are left out.

edkerk:
Thanks for the feedback, it's changed now!

feiranl (approved):
LGTM!

PR #195, opened 2023-01-18
Relax kcats3

johan-gson:
* A bunch of modifications to the GECKO workflow - polished and fixed here and there
* sensitivityTuning added for full and light models.
* The workflow seems to work now for Human-GEM, have not tested for Yeast. Dlkcat remains to fix as well as some other issues such as setting the total pool size, but let's get this in now.


edkerk (approved):
Basic functionality works here. Related aspects indeed still need to be fixed, but this forms a good basis to work from.

PR #198, opened 2023-02-09
fix/feat: various related to kcat gathering, proteomics integration, etc.

edkerk:

 :rotating_light: **After this PR, you should reinstall the DLKcat-GECKO package.**:rotating_light:
Run `system('pipenv rm')` in the dlkcat folder, and remove the dlkcat afterwards. You might have to manually remove any remaining files of the virtual environment at whatever location it is stored on you computer (Windows: `C:\Users\*username*\.virtualenvs\`)

- fix:
  - `findMetSmiles` was unreliable, the web calls sometimes failed which lead to that it was not deterministic and produced different results every time it was run. A loop was added that runs it again if it fails for a metabolite. @johan-gson
  - Modified `writeDLKcatInput` to be able to handle light model as well. @johan-gson
  - corrected stochiometric coefficients of protein usage @johan-gson

- feat:
  - Added function`mergeDlkcatAndFuzzyKcats.m` for merging dlkcat and fuzzy kcats, supporting different priorities depending on wildcard and origin matches in fuzzy. Not tested extensively at this point. @johan-gson 
  - `setProtPoolSize.m` function for setting total protein pool size. @johan-gson
  - `selectKcatValue` can set unique source (model.ec.source) per kcat value, relevant if `mergeDlkcatAndFuzzyKcats` was used.
  - `findMetSmiles` stores model-specific SMILES and queries this first when it runs again
  - `runDLKcat` uses one DLKcat.tsv file, instead of separate in- and output files, to avoid duplication in GECKO model folders. 
  - `makeEcModel` checks that model does not have `_REV` or `_EXP_` in reaction IDs
  - `sensitivityTuning` provides output of modified kcat values, and allows setting by what fold-change the kcat values should be adjusted (new default is 2-fold, instead of 10-fold).
  - `loadDatabases` can use either `proteome` or `taxonomy_id` for Uniprot, as specified in ModelAdapter
  - expanded on `protocol.m`
 
- refactored
  - `sigmaFitter`
  - `calculateFfactor` instead of `getAbundances`


src/geckomat/gather_kcats/readDLKcatOutput.m
```diff
@@ -31,7 +31,7 @@
 params = modelAdapter.params;
 
 if nargin<2 || isempty(outFile)
-    fID      = fopen(fullfile(params.path,'data','DLKcatOutput.tsv'),'r');
+    fID      = fopen(fullfile(params.path,'data','DLKcat.tsv'),'r');
 else
     fID      = fopen(outFile);
 end
```
ae-tafur, line 37: I just have an example file, but I notice that line 38 (textscan) is taking the first line as a header. In the example file (from teams) there is no headers.

edkerk: Yes, I'm about to write this, but with this PR you should redownload and reinstall DLkcat (so run `system('pipenv rm')` from the dlkcat folder and remove the folder afterwards). There are minor changes in the DLKcat I/O files, including this step here.

src/geckomat/gather_kcats/selectKcatValue.m
```diff
@@ -55,18 +55,28 @@
 % Make vector with single kcat value per reaction
 idxInModelUnique = unique(idxInModel);
 selectedKcats    = zeros(numel(idxInModelUnique),1);
+selectedSource   = cell(numel(selectedKcats),1);
+if ~isfield(kcatList,'kcatSource')
+    kcatList.kcatSource = cell(numel(selectedKcats),1);
```
ae-tafur, line 60: `kcatList.kcatSource` should be same size as `kcats` field in `kcatList`. If this is based on` selectedKcats` will get since `selectedKcats` can contain less elements than `kcatList`. 

```suggestion
    kcatList.kcatSource = cell(numel(selectedKcats),1);
```

tutorials/protocol.m
```diff
@@ -68,27 +86,68 @@
...
-% Requires EC-codes
-ecModel_fuzzy   = getECfromGEM(ecModel);
-kcatList        = fuzzyKcatMatching(ecModel_fuzzy);
-ecModel_fuzzy   = selectKcatValue(ecModel_fuzzy, kcatList);
+% Requires EC-codes. Can be either gathered from the model.eccodes field,
+% or from the Uniprot (and KEGG) data
+%ecModel         = getECfromGEM(ecModel);
+ecModel         = getECfromDatabase(ecModel);
+kcatList_fuzzy  = fuzzyKcatMatching(ecModel);
+ecModel_fuzzy   = selectKcatValue(ecModel_fuzzy, kcatList_fuzzy);
```
ae-tafur, line 108: `ecModel_fuzzy` has not been created yet. 

```suggestion
ecModel_fuzzy   = selectKcatValue(ecModel, kcatList_fuzzy);
```

ae-tafur (commented):
Few bugs identified.

src/geckomat/gather_kcats/runDLKcat.m
```diff
@@ -118,18 +111,17 @@ function runDLKcat(DLKcatInput, DLKcatOutput, modelAdapter, DLKcatPath, pythonPa
             [~, pythonPath] = system('which python');
         end
     end
+end
+if endsWith(pythonPath,'.exe')
+    pythonPath = pythonPath(1:end-4);
+end
+if endsWith(pythonPath,'python3')
```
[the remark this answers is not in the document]
edkerk: Will resolve this with an earlier strtrim call in lines 101 and 111.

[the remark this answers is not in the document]
edkerk: No I won't, my solution doesn't work, yours does :)

ae-tafur (approved):

ae-tafur:
I just push a commit correcting and error introduced by GitHub, since does not replace all the lines needed.

PR #199, opened 2023-02-09
Feat: apply kcats defined by the user 

ae-tafur:

A custom Kcat can be defined by user in `data/customKcats.tsv`, changing `model.ec.kcat` value a making an annotation in the model about this, and `applyKcatConstraints`. This change will be done if there is a full match between the custom data and the rules in the model. If a match is >50% & <100%, `applyCustomKcats` will return a table with a list containing the information for potential curation by the user.

`userData/ecYeastGEM/data/manual_data.txt` and `userData/ecYeastGEM/code/manualModifications.m` was incorporated in `userData/ecYeastGEM/data/customKcats.tsv`


edkerk:
Looks good! I made a few changes to:
- allow specifying path to another tsv input file
- allow specifying a matlab structure instead of text file as input
- modify model.ec.source to mention 'custom' if kcat values were changed
- append notes from the input file/structure to model.ec.notes, instead of having a generic text.

I have not tested yet the use of reaction identifiers in the input file. Can we modify the yeast-GEM customKcat.tsv to at least include one of such an example?

edkerk:
Also, GECKO light is not yet implemented, but that can be done after this PR is merged.

ae-tafur:
> Looks good! I made a few changes to:
> 
> * allow specifying path to another tsv input file
> * allow specifying a matlab structure instead of text file as input
> * modify model.ec.source to mention 'custom' if kcat values were changed
> * append notes from the input file/structure to model.ec.notes, instead of having a generic text.
> 
> I have not tested yet the use of reaction identifiers in the input file. Can we modify the yeast-GEM customKcat.tsv to at least include one of such an example?

`customKcats.tsv` file in ecYeastGEM already have some identifiers. Lines 88-89 for `P07149 + P19097`, rxns `r_2140` `r_2041`

edkerk (approved):

PR #200, opened 2023-02-11
Feat/bayesian

johan-gson:
* Added the Bayesian. Note that it is currently limited, it is a bit hardcoded for Yeast-GEM and only works with light models - this is how the code in dlkcat was written. The full model is a bit more complicated to handle, the reactions have different names, sometimes you need to manipulate several reactions, etc., and I have not looked into what needs to be done about that. I have tested to run it once, it took 3-4 hours on my computer (8 + 8 cores), it converged in about 50 iterations - for the full it will take longer. Note that I added three files containing growth data etc. needed to run it.

- [X] Selected `devel3` as a target branch (top left drop-down menu)


johan-gson:
@feiranl Could you test to run it on your computer and see that it does what it should?

edkerk (approved):
With room for improvement, the code seems to work okay. Good to have it combined with the other changes in `devel3`.

feiranl (commented):
Sorry for the late reply. Seems fine.  I was not able to run several lines in the testworkflow. So I manually check these functions.

mihai-sysbio:
I see this PR has been already merged, but @feiranl had some good feedback. Any suggestions @johan-gson @edkerk ?

johan-gson:
This can be done, we just don't have time for it now + we don't have data for other models. Perhaps we could do it for different yeast species, I'm not sure. But time is the major problem, we only have a week until submission, the code should have been frozen by now :).

johan-gson:
Eventually it would be nice to do it for human for example, although I think it may be challenging - I suspect that enzyme availability may not be the limiting factor for growth.

PR #206, opened 2023-02-15
Feat: calculate a standard kcat value for missing GPR reactions

ae-tafur:

A standard kcat value is calculate for those reactions with missing GPR, except transport, spontaneous, exchange, and pseduoreactions.

A standard Kcat is defined for all the rxns which does not have a GPR rule defined. In this case, the kcat value for a particular reaction is defined as the mean of the kcat values of the reactions involved in the same subsystem in which the given reaction is involved. Nevertheless, if a subSystem have a number of reactions lower than a treshold, the kcat value will be the median of the kcat in all the reactions of the model.

Additionally, an stardard MW is defined for all the rxns which does not have a GPR rule defined. This is based in all the proteins reported for the specific organism in uniprot

A pseudometabolite named 'prot_standard' will be added as well as a gene and enzyme named 'standard'. 

Function returns the input needed for `applyKcatConstraints`

### DIFF
--- a/src/geckomat/gather_kcats/getStandardKcat.m
+++ b/src/geckomat/gather_kcats/getStandardKcat.m
@@ -114,6 +114,18 @@ end
 
 model = addGenesRaven(model, proteinStdGenes);
 
+% Add a protein usage reaction
+if ~model.ec.geckoLight
+    proteinStdUsageRxn.rxns            = {'usage_prot_standard'};
+    proteinStdUsageRxn.rxnNames        = proteinStdUsageRxn.rxns;
+    proteinStdUsageRxn.mets         = {'prot_pool', proteinStdMets.mets};
+    proteinStdUsageRxn.stoichCoeffs = [-1, 1];
+    proteinStdUsageRxn.lb              = 0;
+    proteinStdUsageRxn.grRules         = proteinStdGenes.genes;
+    
+    model = addRxns(model, proteinStdUsageRxn);
+end
+
 % Update .ec structure in model
 model.ec.genes(end+1)      = {'standard'};
 model.ec.enzymes(end+1)    = {'standard'};
--- a/src/geckomat/gather_kcats/getStandardKcat.m
+++ b/src/geckomat/gather_kcats/getStandardKcat.m
@@ -7,26 +7,27 @@ function [model, rxnsMissingGPR, standardMW, standardKcat] = getStandardKcat(mod
 %   model           an ecModel in GECKO 3 version
 %   modelAdapter    a loaded model adapter (Optional, will otherwise use the
 %                   default model adapter).
-%   threshold     a threshold to determine when use a kcat value based on
-%                      the mean kcat of the reactions in the same subSystem or 
-%                      based on the median value of all the kcat in the
-%                      model. Second option is used when the number of
-%                      reactions in a determined subSystem is < threshold.
-%                      (default = 10)
+%   threshold       a threshold to determine when use a kcat value based on
+%                   the mean kcat of the reactions in the same subSystem or
+%                   based on the median value of all the kcat in the model.
+%                   Second option is used when the number of reactions in a
+%                   determined subSystem is < threshold. (Optional, default = 10)
 %
 % Output:
-%   model          ecModel where kcats for defined proteins have been
-%                       changed
-%   rxnsMissingGPR   a list of update rxns index with a standard value 
-%   standardMW        the standard MW value calculated
-%   standardKcat       the standard Kcat value calculated 
+%   model           ecModel where model.ec is expanded with a standard
+%                   protein with standard kcat and standard MW, assigned to
+%                   reactions without gene associations.
+%   rxnsMissingGPR  a list of update rxns index with a standard value
+%   standardMW      the standard MW value calculated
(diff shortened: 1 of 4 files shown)


edkerk:
Looks good! I made a few minor tweaks. More significant, I think it should also add a usage reaction (`usage_prot_standard`), similar as is done in step 13 of `makeEcModel`.

Also, once this PR is merged, we should test to run this function immediately after `makeEcModel` and before running any other function, just to make sure that the other functions do not act up because of this change. But that can be after merging.

ae-tafur:
> Looks good! I made a few minor tweaks. More significant, I think it should also add a usage reaction (`usage_prot_standard`), similar as is done in step 13 of `makeEcModel`.
> 

Yeap, you are right. I will do it. 

> Also, once this PR is merged, we should test to run this function immediately after `makeEcModel` and before running any other function, just to make sure that the other functions do not act up because of this change. But that can be after merging.

Just like a test ? or do you mean de order in the protocol ? 

I think this function should be after DLKcat and/or fuzzy kcat in order to get a best standard kcat value.

src/geckomat/gather_kcats/getStandardKcat.m
```diff
@@ -0,0 +1,152 @@
...
+
+% An standard Kcat is defined for all the rxns which does not have a GPR
+% rule defined. In this case, the kcat value for a particular reaction is
+% defined as the mean of the kcat values of the reactions involved in the
+% same subsystem in which the given reaction is involved. Nevertheless, if
+% a subSystem have a number of reactions lower than a treshold, the kcat
+% value will be the median of the kcat in all the reactions of the model.
+
+% Get the kcat value based on all the kcats in the model
+standardKcat = median(model.ec.kcat, 'omitnan');
```
johan-gson, line 59: I'm not sure how common zeros are in the kcat list, but since you check for them below, perhaps they should be omitted in this median calculation as well.

edkerk:
> > Also, once this PR is merged, we should test to run this function immediately after `makeEcModel` and before running any other function, just to make sure that the other functions do not act up because of this change. But that can be after merging.
> 
> Just like a test ? or do you mean de order in the protocol ?
> 
> I think this function should be after DLKcat and/or fuzzy kcat in order to get a best standard kcat value.

Yes, sorry, I indeed meant just like a test. Intuitively one should indeed run this late in the pipeline, but the user might want to go back and rerun some earlier step. Maybe writeDLKcatInput will be confused that there is no sequence for the last enzyme for instance.

johan-gson (approved):
Looks very good, nice code! I just had one tiny comment. One thing to think about is if it should be mean or median - you do median for the standard value and mean within each group - I'm not sure if there is a thought behind that (then forget this comment) or if it just happened. The reason why I would prefer median is that it is more robust for long-tailed distributions such as the log-normal we see for kcats.

src/geckomat/gather_kcats/getStandardKcat.m
edkerk: They should indeed be ignored here, no matter how (un)common they are.

ae-tafur: few zeros, but I will include this

ae-tafur:
> > > Also, once this PR is merged, we should test to run this function immediately after `makeEcModel` and before running any other function, just to make sure that the other functions do not act up because of this change. But that can be after merging.
> > 
> > 
> > Just like a test ? or do you mean de order in the protocol ?
> > I think this function should be after DLKcat and/or fuzzy kcat in order to get a best standard kcat value.
> 
> Yes, sorry, I indeed meant just like a test. Intuitively one should indeed run this late in the pipeline, but the user might want to go back and rerun some earlier step. Maybe writeDLKcatInput will be confused that there is no sequence for the last enzyme for instance.

Absolutely

ae-tafur:
> Looks very good, nice code! I just had one tiny comment. One thing to think about is if it should be mean or median - you do median for the standard value and mean within each group - I'm not sure if there is a thought behind that (then forget this comment) or if it just happened. The reason why I would prefer median is that it is more robust for long-tailed distributions such as the log-normal we see for kcats.

Precisely, because of that I used median for the standard since I have a large amount of data. But within group, I have less values so, it will potentially have a big number. I tested that, so, as result I prefer to use mean with in groups

johan-gson:
Ok, nice, I will add this function to the verification matrix, could you by any chance write a test case in a similar style as the other testcases? I think you could just create a nonsense model with nonsense reactions and two subsystems, where you just assign some kcats and some gaps and check if they are filled in properly. If you need an adapter you can probably use the TestAdapter.

ae-tafur:
> Ok, nice, I will add this function to the verification matrix, could you by any chance write a test case in a similar style as the other testcases? I think you could just create a nonsense model with nonsense reactions and two subsystems, where you just assign some kcats and some gaps and check if they are filled in properly. If you need an adapter you can probably use the TestAdapter.

Sure. Although, @edkerk  suggest to test after merge. So, we merge first or I add the test before ?

johan-gson:
To me before or after doesn't matter much, just that it gets done :)

edkerk (approved):

Issue #205, opened 2023-02-15
Error applyKcatConstrain


ae-tafur:
I identified an error here:

https://github.com/SysBioChalmers/GECKO/blob/12aeddbee0a48bfe9ebe84c2d3ca621549bc29ac/tutorials/protocol.m#L115

What happen ? 

when `printRxnFormula(model,'r_0226')` this is the output (r_0226 is the ATP synthase mitochondrial reaction):

```
at line 56:
	r_0226	s_0397 + 3 s_0794 + s_1326 	->	s_0437 + 2 s_0799 + s_0807

at line 74:
	r_0226	s_0397 + 3 s_0794 + s_1326 	->	s_0437 + 2 s_0799 + s_0807

at line 99:
	r_0226	s_0397 + 3 s_0794 + s_1326 	->	s_0437 + 2 s_0799 + s_0807
```
At this point is ok, since not kcat constrains have not been applied. 

When `applyKcatConstraints` (line 100, see output below) to the ecModel based on the previous kcatList_DLKcat, not change is observed. But, it is ok, since `r_0226` is not present in the list.

```
at line 100 (DLKcat):
	r_0226	s_0397 + 3 s_0794 + s_1326 	->	s_0437 + 2 s_0799 + s_0807
```

When `applyKcatConstraints` under fuzzy matching (output below), it is observed the expected result. Here, it is ok, since in kcatList_fuzzy it is present 'r_0226'.
```
at line 109 (fuzzy):
	r_0226	s_0397 + 3 s_0794 + s_1326 + 0.18338 prot_P00830 + 0.0973862 prot_P00854 + 0.0194846 prot_P00856 + 0.0901104 prot_P05626 + 0.196145 prot_P07251 + 0.229056 prot_P09457 + 0.0225669 prot_P21306 + 0.0662985 prot_P30902 + 0.114963 prot_P38077 + 0.0259672 prot_P61829 + 0.109197 prot_P81449 + 0.0223829 prot_P81450 + 0.0252142 prot_P81451 + 0.0378581 prot_Q06405 + 0.0569612 prot_Q12165 + 0.43243 prot_Q12233 + 0.0472825 prot_Q12349 	->	s_0437 + 2 s_0799 + s_0807
```
But when merging, the output will depend of the model you set as input 

```
at line 115 (merged):
	

> if uses ecModel from line 100 in selectKcatValue:

        ecModel_merged  = selectKcatValue(ecModel, kcatList_merged);
[...]

PR #207, opened 2023-02-18
feat: various relatively minor feats and fixes

edkerk:
- feat:
  - wrapper IO functions `loadEcModel`, `saveEcModel`, `loadConventionalGEM`, that can take modelAdapter info as input, to simplify loading and saving of models. It's just a wrapper for `readYAMLmodel`, `writeYAMLmodel` and `importModel`, so these can still be used as well.
  - `getKcatAcrossIsoenzymes` assigns mean kcat to kcat-less reactions that have isoenzymes with kcat values.
  - `loadDatabases` can output alternative gene annnotations instead of KEGG gene IDs. `kegg.tsv` therefore has an additional column, to ensure that kegg gene ID is still present. Also, `uniprot.tsv` has been slightly shrunk by removing unused gene names annotation.
  - `applyCustomKcats` also allows providing only the original reaction ID (without the _EXP_ specification) and no protein IDs. In that case, it will apply the same kcat to all isoenzymes.

- fix:
  - `mergeDLKcatAndFuzzyKcats` changed default parameters, to solve #205. Also documentation slightly extended to exemplify the meaning of the parameters.

- chore:
  - remove deprecated files and functions


src/geckomat/change_model/getKcatAcrossIsoenzymes.m
```diff
@@ -0,0 +1,38 @@
+function model = getKcatAcrossIsoenzymes(model)
+% getKcatAcrossIsoenzymes
```
johan-gson, line 2: I guess this is only relevant for full ec models? May be good to write in the text and check in the function.

johan-gson: Will you add a test case for this function? It should be relatively easy - you could copy parts of tc0011 and just zero the kcat for reaction R2_EXP_2 or something like that. And then make sure it gets filled in from R2_EXP_1. Make sure it works also for REV.

ae-tafur (approved):
Looks good! 

Should we include here as deprecated `utilities/changeMedia_yeast.m`, or move this to `userData/ecYeastGEM/code` ?

src/geckomat/change_model/getKcatAcrossIsoenzymes.m
edkerk: Or would you want something similar for light models? Then the smallest isoenzyme will always be selected, while if that isoenzyme would not have had a kcat value it would never be active? (I assume that light ignores isoenzymes with 0 kcat value?)

edkerk:
> Looks good!
> 
> Should we include here as deprecated `utilities/changeMedia_yeast.m`, or move this to `userData/ecYeastGEM/code` ?

Good suggestion, will have to be refactored later.

src/geckomat/utilities/saveEcModel.m
```diff
@@ -0,0 +1,69 @@
+function saveECmodel(model,modelAdapter,format,filename)
```
johan-gson, line 1: Could you update tc0009 to test these three functions, i.e., load a conventional GEM, save it as ec, load it, and compare? Should be easy. Ideally the model should go through the processing chain so we test that metSmiles etc. are saved.

johan-gson (approved):
I have some comments, mostly about testing. At this point, I think it would be good to add test cases for the new functions we add - we need to test them at some point anyway.

johan-gson:
Another comment I forgot, about Human-GEM. I got the feeling that the model is stored in GECKO, is that so? Human-GEM still needs to be installed, we use gene conversions etc. from that installation - isn't it better to load the model from it's real location?

src/geckomat/change_model/getKcatAcrossIsoenzymes.m
johan-gson: Yeah, they are ignored, but then it would be the mean of the other possibilities, and it would not be the smallest, and not be used :). I don't think we need one here - it is only if you constrain individual enzymes that this makes sense unless I misunderstood something?

edkerk: Right, indeed, I'll then just let it throw an error if applied to geckoLight model.

edkerk:
> Another comment I forgot, about Human-GEM. I got the feeling that the model is stored in GECKO, is that so? Human-GEM still needs to be installed, we use gene conversions etc. from that installation - isn't it better to load the model from it's real location?

Probably. The model is currently not in GECKO. I've changed the adapter to refer to the Human-GEM location, similar as is done for getting Uniprot compatible genes

PR #215, opened 2023-02-20
Feat: flexibilize concentrations of proteins after proteomics integration

ae-tafur:

-  Update the `proteomics.tsv` files with the correct units `ug/gDW`
- `proteomics.tsv` file can store multiple conditions. Then, `readProteomcs` was refactored. 
- flexibilize proteins concentrations after proteomics integration to reach an experimental growth rate. A loop is performed until the specified growth rate is reached. Protein usage with the highest impact over the growth rate is selected to be increased. However, is was observed that same enzyme can be predicted many times to be increased. In this case, a particular enzyme can be increased up to five times, then loop ends, generating an error which suggest manual curation (adjust kcat) of that enzyme. It can be done with `applyCustomKcats`  

test: to include


johan-gson (approved):
Looks reasonable, I just didn't have time to go through it in super detail. But it seems the strategy is very similar as the simple kcat relaxation we have. Good that you have implemented this!

johan-gson:
Will you also add a test case? It is not super easy to do, not sure we have to, I have not done it for the kcat relaxation.

edkerk:
Looks good, but one thing to consider is that a 5-times increase by 10% means a maximum of 161% before the function stops, that doesn't seem like a lot to me. With a default `foldChange` of 1.25, it already allows a maximum of ~300% of measured enzyme level, which doesn't seem extreme? Of course the user can define `foldChange`, but perhaps the user should also be able to define how many rounds of increases are allowed.

Somewhat related to that, I think there should be an option to not have any limit on the number of increases: that way the function can run on many models in batch mode, at least yielding a functional model, and the user can then decide afterwards if they want to change something. Currently, the function would break and prevent any further work with the model.

So, a 'iterationsPerEnzyme' flag that can be numeric (with default 5? or maybe 10 for allowing [enzyme] to be 1.1^10= 250% of measured level, or Inf, for unlimited?

ae-tafur:
> Will you also add a test case? It is not super easy to do, not sure we have to, I have not done it for the kcat relaxation.

Yes, I can do it

ae-tafur:
> Looks good, but one thing to consider is that a 5-times increase by 10% means a maximum of 161% before the function stops, that doesn't seem like a lot to me. With a default `foldChange` of 1.25, it already allows a maximum of ~300% of measured enzyme level, which doesn't seem extreme? Of course the user can define `foldChange`, but perhaps the user should also be able to define how many rounds of increases are allowed.
> 
> Somewhat related to that, I think there should be an option to not have any limit on the number of increases: that way the function can run on many models in batch mode, at least yielding a functional model, and the user can then decide afterwards if they want to change something. Currently, the function would break and prevent any further work with the model.
> 
> So, a 'iterationsPerEnzyme' flag that can be numeric (with default 5? or maybe 10 for allowing [enzyme] to be 1.1^10= 250% of measured level, or Inf, for unlimited?

It seems reasonable to used 10 rounds, I tested with yeast, and usually it predict changes in low concentrations (a limitation of proteomics) and enzymes with low kcat.

ae-tafur:
test case for proteomics is added now (tc0013). 

fix a bug getting standard Kcat when there is more than one subSystem defined for the reaction

edkerk:
Looks good, let's merge to put it all together.

Issue #214, opened 2023-02-20
Bug in getECfromDatabase.m


Yu-sysbio:
https://github.com/SysBioChalmers/GECKO/blob/ee76b14e767e658b1f3a88ce70732ecdca3f0884/src/geckomat/get_enzyme_data/getECfromDatabase.m#L80

If KEGG is skipped, i.e., Line 53 is false, then `DBecNumKEGG`, `DBMWKEGG`, `geneIndexKEGG` and `geneHashMapKEGG` would be unrecognized.

Issue #210, opened 2023-02-19
genes not found in local UniProt DB (minor)


Yu-sysbio:
Potential minor enhancement for your consideration

https://github.com/SysBioChalmers/GECKO/blob/e97f0e69307582048b2776aa74eab3e98156c694/src/geckomat/change_model/makeEcModel.m#L241

Is it necessary to display (maybe just in the command window) those genes that cannot be found in local UniProt DB and thus not be enzyme-constrained?
Sometimes I am quite curious which genes they are and intend to make them constrained. Or we could tell a solution in the protocol, e.g., compare ecModel.genes vs ecModel.ec.genes.

edkerk:
Good point, I'm actually planning to make it an additional output. It might also be helpful to decide which Uniprot dataset (by taxonomy id or proteome id) to use.

PR #218, opened 2023-02-21
fix: various tweaks when going through protocol

edkerk:

- `findMetSmiles` do not override existing metSmiles by default (model might have more than the function can find).
- allow providing of no keggID or keggGeneIdentifier (solves #214)
- `makeEcModel` outputs non-matching genes (solves #210)
- tests for model input/output and applyKcatAcrossIsoenzymes
- `writeDLKcatInput` prevent overwrite of existing file, to avoid loss of kcat values
- fix `flexibilizeProtConcs` and speed-up `getConcControlCoeff` with hot-start
- updated `protocol.m`


ae-tafur (approved):
Should we also deprecate limitingUBs ?

src/geckomat/limit_proteins/flexibilizeProtConcs.m
```diff
@@ -57,7 +57,7 @@
 predGrowth = abs(sol.f);
 
 % Get those proteins with a concentration defined
-protConcs = ~isnan(model.ec.concs);
+protConcs = find(~isnan(model.ec.concs));
```
ae-tafur, line 60: I think that using a logical array is faster than using find after get it. But here I don't think is extreme the time increased

edkerk: It has to do with the change in line 90, which requires it to be indices not logical.

edkerk:
> Should we also deprecate limitingUBs ?

Yes, this is completely replaced with your recent code.

PR #225, opened 2023-02-21
feat: GH Action to run tests

### LINKED ISSUE #223: GitHub Action that runs core function tests with each PR
### Description of the new feature:
Similar as is done for instance in [RAVEN](https://github.com/SysBioChalmers/RAVEN/blob/main/.github/workflows/testing.yml), there should be a GitHub Action that runs unit tests with each PR.

The code should:
- [x] Clone the latest RAVEN release & run `checkInstallation` (it comes with glpk solver, which is enough for the tests, so no further settings are required).
- [x] Run `GECKOInstaller.install` from the GECKO main root.
- [x] Run `runtests('geckoCoreFunctionTests')`.
- [x] Add the output from `runtests` as a comment in the PR.
  - The output from `checkInstallation` does not need to be shown.

mihai-sysbio:

This PR resolves #223 by introducing a GH Action to run the tests. Also, it adds a badge in the README with the status of the tests for the `main` branch, in anticipation that it will be renamed from the current `master`.


mihai-sysbio:
Sorry for the many force-pushes, I really had to hack my way around having the output on multiple lines.

mihai-sysbio:
Also, just stating the obvious: due to how the tests run, even when getting errors, tests are passing.

edkerk:
Hehe, no problem. But it seems to clone RAVEN 2.6.2, instead of the latest release (2.7.12). This is most likely why all these tests fail.

edkerk:
I see from your last commit what might be the problem. In MATLAB you should then remove all RAVEN subfolders. Since version 2.7.9 there is a `removeRavenFromPath` function. If you first `cd('installation')` and then run `removeRavenFromPath`, it _should_ also remove other recent RAVEN paths.

mihai-sysbio:
Thanks for the tip @edkerk! It seems to attempt to remove the RAVEN paths, but it ends up failing with:
```
Error using ls (line 47)
ls: cannot access
'/home/m/actions-runner/_work/GE ..... ' % a huge string here
File name too long
Error in removeRavenFromPath (line 23)
    dirCont = ls(currPath{i});
```

edkerk:
In that case you might have to "manually" ignore the subfolders that are not in the path. Perhaps it is useful to look at how these folders (`.git` etc.) were ignored when `checkInstallation` added them to MATLAB path:

https://github.com/SysBioChalmers/RAVEN/blob/6638c2bab213e143fe690e6f05d7a1e9ce953194/installation/checkInstallation.m#L69

edkerk:
I'll fix `checkInstallation`, at least so not to throw an error if `makeBinaryExecutable` failes, so that the rest of the function can continue (and the binaries are not used in the functions here anyway).

As work-around you can add all RAVEN (sub)folders to the RAVEN path and skip `checkInstallation`. Instead, run `setRavenSolver('glpk')` (or 'gurobi' if available), I think that should cover all GECKO functions.

mihai-sysbio:
> I'll fix RAVEN's `checkInstallation`, at least so not to throw an error if `makeBinaryExecutable` failes, so that the rest of the function can continue (and the binaries are not used in the functions here anyway).

It was driving me nust, the `which` command fails for _some_ files in the `RAVEN` folder. And I finally understood what's going on: the `which` command fails on this Linux Mint machine if the file has no extension.

mihai-sysbio:
@edkerk I though [this fix](https://github.com/SysBioChalmers/RAVEN/commit/6f71555b7df8f68c3ae5e4d48b6fd9d55f381add) would work, but apparently not 😕

edkerk:
Right, we can test there.

mihai-sysbio:
This PR will be ready for review as soon as the changes in [this RAVEN PR](https://github.com/SysBioChalmers/RAVEN/pull/468) make it to `main`.

edkerk:
As a temporary solution, it would be okay to skip `checkInstallation`, as the most important aspect of that function for GECKO is that RAVEN is added to the MATLAB path, and that you already do with `addpath(genpath('RAVEN'))`. Reviewing that RAVEN PR + making a release has lower priority right now, while it would be good to have these tests already running here on the GECKO repo.

edkerk:
It prints the output very nicely, ready to push?

mihai-sysbio:
> It prints the output very nicely, ready to push?

I'd say so - but then, it looks like there's an error caught by the tests, so that ought to be reported somewhere and fixed.

edkerk (approved):
Failed tests are fixed

edkerk:
I clicked to fast to choose "squash"...

PR #226, opened 2023-02-22
feat: run DLKcat from a Docker image

ae-tafur:

- Uses a Docker image to run DLKcat in cases `runDLKcat` fails. I test it on my computer so, It would be great if several of you could test it on other computers (MacOS, Windows, and different architecture).

Download the DLKcat can take some time depending of course of the internet speed. It is around 1.9 GB

Docker desktop can be dowloaded from "https://www.docker.com". When running DLKcat docker CLI needs Docker desktop running.

### DIFF
--- a/src/geckomat/gather_kcats/runDLKcatFromDocker.m
+++ b/src/geckomat/gather_kcats/runDLKcatFromDocker.m
@@ -89,7 +89,7 @@ if checks.image.status == 0 && ~contains(checks.image.out,'dlkcat')
         error('Fail to load the image')
     end
 elseif checks.image.status ~= 0
-    error('Docker is installed. Check if the docker desktop app is running')
+    error('Looks like Docker is installed, but make ensure that the docker desktop app is running.')
 end
 
 % Copy the input file for needed for the docker image to run DLKcat
--- a/src/geckomat/gather_kcats/runDLKcatFromDocker.m
+++ b/src/geckomat/gather_kcats/runDLKcatFromDocker.m
@@ -77,7 +77,7 @@ cd(DLKcatPath);
 if checks.image.status == 0 && ~contains(checks.image.out,'dlkcat')
     % Check if the image file is already dowloaded
     if ~exist(fullfile(DLKcatPath,'dlkcat_docker.tar.gz'),'file')
-        disp('=== Downloading DLKcat docker image, this may take several minutes...')
+        disp('=== Downloading DLKcat docker image, this may take more than 20 minutes, depending on your internet connection speed...')
         packageURL = 'https://github.com/SysBioChalmers/GECKO/raw/dlkcatPackage/dlkcat_docker.tar.gz';
         websave(fullfile(DLKcatPath,'dlkcat_docker.tar.gz'),packageURL);
     end
--- a/src/geckomat/gather_kcats/runDLKcatFromDocker.m
+++ b/src/geckomat/gather_kcats/runDLKcatFromDocker.m
@@ -47,7 +47,7 @@ end
 % On Mac, docker might not be properly loaded if MATLAB is started via
 % launcher and not terminal.
 if ismac
-    setenv('PATH', strcat('/usr/local/bin/', ':', getenv("PATH")));
+    setenv('PATH', strcat('/usr/local/bin', ':', getenv("PATH")));
 end
 
 % Check if docker is installed
--- a/src/geckomat/gather_kcats/runDLKcatFromDocker.m
+++ b/src/geckomat/gather_kcats/runDLKcatFromDocker.m
@@ -1,0 +1,111 @@
+function runDLKcatFromDocker(deleteImage, DLKcatFile, modelAdapter, DLKcatPath)
+% runDLKcatFromDocker
+%   Runs DLKcat to predict kcat values from a Docker image
+%
+% Input
+%   deleteImage      true or false if delete the docker image tar.gz file.
+%                   (Optional, default = false)
+%   DLKcatFile      path to the DLKcat.tsv file (including file name), as
+%                   written by writeDLKcatFile. Once DLKcat is succesfully
+%                   run, the DLKcatFile will be overwritten with the DLKcat
+%                   output. Optional, otherwise the file location will be
+%                   assumed to be in the model-specific 'data' sub-folder
+%                   taken from modelAdapter (e.g.
+%                   GECKO/userData/ecYeastGEM/data/DLKcat.tsv)
+%   modelAdapter    a loaded model adapter. (Optional, will otherwise use
+%                   the default model adapter)
+%   DLKcatPath      path where DLKcat is/will be installed. (Optional,
+%                   defaults to GECKO/dlkcat)
+%
+%   NOTE: 1. Requires Docker to be installed. Visit "https://www.docker.com"
+%         2. Runtime will depend on whether the image is to be downloaded or not.
+
+%Get the GECKO path
+geckoPath = findGECKOroot();
+
+if nargin < 4 || isempty(DLKcatPath)
+    DLKcatPath = fullfile(geckoPath,'dlkcat');
+end
+
+if nargin < 3 || isempty(modelAdapter)
+    modelAdapter = ModelAdapterManager.getDefaultAdapter();
+    if isempty(modelAdapter)
+        error('Either send in a modelAdapter or set the default model adapter in the ModelAdapterManager.')
+    end
+end
+params = modelAdapter.params;
+
+if nargin < 2 || isempty(DLKcatFile)
+    DLKcatFile = fullfile(params.path,'data','DLKcat.tsv');
+end
+
+if nargin < 1 || isempty(DLKcatPath)
+    deleteImage = false;
+end
+
+%% Check and install requirements
+% On Mac, docker might not be properly loaded if MATLAB is started via
+% launcher and not terminal.
+if ismac
+    setenv('PATH', strcat('/usr/local/bin/', ':', getenv("PATH")));
+end
+
+% Check if docker is installed
+[checks.docker.status, checks.docker.out] = system('docker --version');
+if checks.docker.status ~= 0
+    error('Cannot find Docker. Make sure it is installed')
+end
+
+% Dowload DLKcat package
+if ~exist(fullfile(DLKcatPath,'DLKcat.py'),'file')
+    if ~exist(fullfile(DLKcatPath),'dir')
+        mkdir(fullfile(DLKcatPath));
+    end
+    disp('=== Downloading DLKcat ...')
+    packageURL = 'https://github.com/SysBioChalmers/GECKO/raw/dlkcatPackage/dlkcat.zip';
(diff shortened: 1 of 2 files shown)


ae-tafur:
Last commit add at the environment the path where usually the libraries on MacOS are. Since I notice most of the issues here are related to MacOS

mihai-sysbio:
A general comment regarding all the paths that are set in ` src/geckomat/gather_kcats/runDLKcat.m`: to me, this setting of paths belongs better in the installation part of GECKO.

src/geckomat/gather_kcats/runDLKcatFromDocker.m
```diff
@@ -0,0 +1,111 @@
...
+    [checks.dockerload.status, checks.dockerload.out] = system('docker load --input dlkcat_docker.tar.gz');
+    if deleteImage
+        delete(fullfile(DLKcatPath,'dlkcat_docker.tar.gz'));
+    end
+    if checks.dockerload.status ~= 0
+        error('Fail to load the image')
+    end
+elseif checks.image.status ~= 0
+    error('Docker is installed. Check if the docker desktop app is running')
+end
```
mihai-sysbio, line 93: If the image were to be published, most of this code would not be needed. So that is what I would strongly advise, especially since it would bring transparency to how the image was created.

src/geckomat/gather_kcats/runDLKcatFromDocker.m
```diff
@@ -0,0 +1,111 @@
...
+
+% Copy the input file for needed for the docker image to run DLKcat
+copyfile(DLKcatFile,fullfile(DLKcatPath, 'data/DLKcat.tsv'));
+
+disp('=== Running DLKcat prediction, this may take several minutes...')
+% In the next line, pythonPath does not need to be specified, because it is
+% already mentioned when building the virtualenv.
+dlkcat.status = system('docker compose -f docker-compose.yaml up dlkcat-cont', '-echo');
```
mihai-sysbio, line 101: Based on how I've used Docker images before, I have a feeling we could do without `docker-compose`, and use instead `docker run`. To be sure, I wanted to look into the `docker-compose.yaml` file, but I couldn't seem to find it.

ae-tafur: Yo can do it, but the problem is that runnnig DLKCat needs a few files that defined at the moment that is running. 

Compose file is in the DLKcat folder

mihai-sysbio: GECKO on branch `feat/runDLKcatFromDocker `
❯ find . -name "*.yaml"

yields no results

@ae-tafur could you please point me directly to the file?

src/geckomat/gather_kcats/runDLKcatFromDocker.m
mihai-sysbio: I'm guessing it's a private personal repository. If it's a pure `DLKcat` image, I guess it would make sense to have it in the [`DLKcat` repository](https://github.com/SysBioChalmers/DLKcat/).

src/geckomat/gather_kcats/runDLKcatFromDocker.m
mihai-sysbio: > needs a few files defined at the moment that is running.

This is routinely done with [`bind` mounts](https://docs.docker.com/storage/bind-mounts/#start-a-container-with-a-bind-mount), so it might be possible to not have a `docker-compose` file.

src/geckomat/gather_kcats/runDLKcatFromDocker.m
edkerk: It contains the minimum code and data to predict kcat values, while the code in DLKcat repo can also regenerate the underlying DL model. Moreover, the `dlkcat.py` script here is modified to use the GECKO input file (it for instance has an extra column with reaction identifiers which are then also included in the output file, and finding SMILES is disabled as it is done in MATLAB instead).

But perhaps relevant to mention is that the `dlkcat.zip` and `DLKcat` image are stored in a separate orphan branch in GECKO GitHub.

src/geckomat/gather_kcats/runDLKcatFromDocker.m
ae-tafur: https://github.com/SysBioChalmers/GECKO/blob/2a9b890cd0495ff9a532a165e3b00ba6a875ec21/dlkcat.zip

src/geckomat/gather_kcats/runDLKcatFromDocker.m
mihai-sysbio: > `dlkcat.zip` and `DLKcat` image are stored in a separate orphan branch in GECKO GitHub.

I think we ought to do better than that. In terms of the image itself, it should be [stored on GH and produced with an action at every release](https://docs.github.com/en/actions/publishing-packages/publishing-docker-images#publishing-images-to-github-packages). @ae-tafur can I give this a go with your code? And then this branch can get some more commits to simplify the code.

edkerk: I also see DLKcat as more of a plugin to GECKO. That's why I would also not already set the paths when running GECKOInstaller.Install.

Yu-sysbio:
Me as a user give feedback below:
1. Before using runDLKcatFromDocker, it would be better to inform that Docker should be installed and ensure that the docker desktop app is running.
2. It took one hour to download DLKcat docker image, which is ok, but it would be better to inform this long time, not just several minutes as displayed in the command window.

Yu-sysbio:
I got the error below:

> Error using copyfile
> cp: /Users/cheyu/Documents/GitHub/GECKO/dlkcat/data/DLKcat.tsv: No such file or directory
> 
> Error in runDLKcatFromDocker (line 96)
> copyfile(DLKcatFile,fullfile(DLKcatPath, 'data/DLKcat.tsv'));

This is because I have the `DLKcat.tsv` saved in the path `GECKO/userData/yourModel/data` following the protocol.m file, which is not the default `fullfile(DLKcatPath, 'data/DLKcat.tsv')`.

ae-tafur:
> Me as a user give feedback below:
> 
> 1. Before using runDLKcatFromDocker, it would be better to inform that Docker should be installed and ensure that the docker desktop app is running.

I mention in the description of the function about getting Docker installed. But it should be also, mention that docker desktop is running. 

We should also include this in the protocol, 

> 2. It took one hour to download DLKcat docker image, which is ok, but it would be better to inform this long time, not just several minutes as displayed in the command window.

I agree, several minutes sound like maximum 5-10 minutes

src/geckomat/gather_kcats/runDLKcatFromDocker.m
```diff
@@ -0,0 +1,111 @@
...
+%                   output. Optional, otherwise the file location will be
+%                   assumed to be in the model-specific 'data' sub-folder
+%                   taken from modelAdapter (e.g.
+%                   GECKO/userData/ecYeastGEM/data/DLKcat.tsv)
+%   modelAdapter    a loaded model adapter. (Optional, will otherwise use
+%                   the default model adapter)
+%   DLKcatPath      path where DLKcat is/will be installed. (Optional,
+%                   defaults to GECKO/dlkcat)
+%
+%   NOTE: 1. Requires Docker to be installed. Visit "https://www.docker.com"
```
ae-tafur, line 20:
```suggestion
%   NOTE: 1. Requires Docker to be installed, and docker desktop running. Visit "https://www.docker.com"
```

src/geckomat/gather_kcats/runDLKcatFromDocker.m
```diff
@@ -0,0 +1,111 @@
...
+    end
+    [checks.dockerload.status, checks.dockerload.out] = system('docker load --input dlkcat_docker.tar.gz');
+    if deleteImage
+        delete(fullfile(DLKcatPath,'dlkcat_docker.tar.gz'));
+    end
+    if checks.dockerload.status ~= 0
+        error('Fail to load the image')
+    end
+elseif checks.image.status ~= 0
+    error('Docker is installed. Check if the docker desktop app is running')
```
ae-tafur, line 92:
```suggestion
    error('Looks like Docker is installed, but make ensure that the docker desktop app is running.')
```

src/geckomat/gather_kcats/runDLKcatFromDocker.m
```diff
@@ -0,0 +1,111 @@
...
+
+currPath = pwd();
+cd(DLKcatPath);
+
+% First check if the Docker image exist
+[checks.image.status, checks.image.out] = system('docker images');
+if checks.image.status == 0 && ~contains(checks.image.out,'dlkcat')
+    % Check if the image file is already dowloaded
+    if ~exist(fullfile(DLKcatPath,'dlkcat_docker.tar.gz'),'file')
+        disp('=== Downloading DLKcat docker image, this may take several minutes...')
```
ae-tafur, line 80:
```suggestion
        disp('=== Downloading DLKcat docker image, this may take more than 20 minutes, depending on your internet connection speed...')
```

Yu-sysbio:
> 1. Before run this, you set the default modelAdapter to be `yourModelAdapter` ?

Yes. I just followed the protocol.m file.

> 2. You download again the DLKcat folder ? I update yesterday the `DLKcat.zip` file which include a folder named `dklcat/data` where the file stored in `GECKO/userData/yourModel/data` (in your case) will be copied

I did not download the DLKcat folder myself but just ran `runDLKcatFromDocker()` once, this can automatically download the DLKcat folder I think.

Maybe you can run the protocol.m to test a bit?

ae-tafur:
> I did not download the DLKcat folder myself but just ran `runDLKcatFromDocker()` once, this can automatically download the DLKcat folder I think.

`runDLKcat` and `runDLKcatFromDocker` does not download the DLKcat package if there is file named `DLKcat.py`. So if you have the old DLKcat.zip, this will no be updated. A solution can be, go to `GECKO/dlkcat` and check if exits a `GECKO/dlkcat/data` dir inside, if not, you can create and run again `runDLKcatFromDocker`. Future users will not need to do this since when you clone the repo and run for the first time you will get the DLKcat package updated

> Maybe you can run the protocol.m to test a bit?

will do It with this branch

Yu-sysbio:
> A solution can be, go to `GECKO/dlkcat` and check if exits a `GECKO/dlkcat/data` dir inside, if not, you can create and run again `runDLKcatFromDocker`.

Following your suggested solution (i.e., manually create `data` folder in the path `GECKO/dlkcat`) (BTW can we do this automatically?), I addressed my previous error, but then I got this:

> stat /Users/cheyu/Documents/GitHub/GECKO/dlkcat/docker-compose.yaml: no such file or directory
> Error using runDLKcatFromDocker (line 108)
> DLKcat encountered an error.

What is the solution then?

ae-tafur:
> 
> What is the solution then?
> 

Yeap, the general problem is that you have the old DLKcat package. Then what do you need to do is delete the GECKO/dlkcat and run again the code it will download all the files needed. 

This is caused because you don't have the updated package. And download it is ignored if matlab find a file named "DLKcat.py"

edkerk (approved):
Also works here!

Issue #239, opened 2023-02-25
Bug detected in findMetSmiles for reconstructing E. coli ecModel


Yu-sysbio:
Reported by @feiranl and I also reproduced the same issue, see below:

Using iML1515 as the starting model, when running to the step `ecModel = findMetSmiles(ecModel)`, it increases to 25% and then stays there -- no error occurs but this number never increases.

> Check for local SMILES database... not found.
> Querying PubChem for SMILES by metabolite names...  25% complete

edkerk:
Can you attach the ecModel as .mat file here, to be able to quickly reproduce the problem?

edkerk:
I cannot replicate this. I think this should only hang if there is problems with the internet connection (it will just keep on retrying). Could you retry, and if it hangs again, press the _Pause_ button, and when paused, click on the _Step Out_ button until you're in the `findMetSmiles` function. Then, run the function step-wise with the _Step_ button and see if it indeed hangs when running `webread`?

feiranl:
I tried, hangs again, and it paused at line 64 `catch exception`, still part of the webread. But if you run that successfully, maybe it is our network connection from Shenzhen. Is that possible to give an instruction to user how to run that themselves beyond the GECKO part?

edkerk:
It involves an API call for each metabolite, so not that straightforward to instruct how to download this manually. But currently `smilesDB.tsv` is only written once the function is done, instead it could write a new line after each succesful call, which would at least avoid having to restart from scratch. I'll refactor a bit and let you know when I have something to test.

Yu-sysbio:
@edkerk Could you check if this file is the same as what you obtained?
[smilesDB.tsv.zip](https://github.com/SysBioChalmers/GECKO/files/10835306/smilesDB.tsv.zip)

edkerk:
No, that file is incorrect. But I've now managed to replicate the problem and will push another fix.

Issue #235, opened 2023-02-24
Outdated output files


ae-tafur:
Data output files in `userDat/ecYeastGEM/output`does not correspond to the model generated in version 3 or they are not longer generated (e.g. txt files). Then those need to be removed and save the new ones. 

- [ ] Done this analysis in the `devel3` branch of the repository
- [ ] Checked that a similar issue does not exist already

PR #245, opened 2023-02-27
feat: simulations

edkerk:

- `mapRxnsToConv` can map a flux solution (e.g. from `solveLP`) from an ecModel to model.rxns order, combining isoenzymes and reversible reactions
- `ecFVA` performs FVA and returns the solution with `mapRxnsToConv`
- `enzymeUsage` can give enzyme absolute and capacity usage based on simulation result
- removed some deprecated functions, moved some yeast-specific functions into its userData
- solves #246, #239, #235, #250


src/geckomat/utilities/enzymeUsage.m
```diff
@@ -0,0 +1,49 @@
+function [absUsage,capUsage,UB,protID] = enzymeUsage(ecModel,fluxes,zero)
```
johan-gson, line 1: Do these functions support the light models? If not, at least write that in the header. It shouldn't be that hard to implement I guess.

johan-gson (approved):
I only looked at briefly at this, just don't have the time right now. But it looks good!

src/geckomat/utilities/enzymeUsage.m
edkerk: Not supported, as it currently relies on the usage_prot reactions. I'll mention this for the moment, and in the long run it can indeed be implemented for geckoLight as well.

PR #247, opened 2023-02-27
fix: protocol updates

### LINKED ISSUE #222: fix: inconsistencies in `userData` folder
### Description of the bug:

This issue is meant to cover all fixes for the `userData` folder that are not related to code, as they are in the `devel3` branch:
- [x] In each userData subfolder, all folders (code, data, models, output) should be present
- [x] the Readmes should be written in markdown format
- [x] there should be only 1 readme for each subfolder in `userData`, e.g. the file `userData/ecHumanGEM/data/Data files go here.txt` should not exist, and instructions for how to use this should be either in the tutorial or in a new file `userData/ecHumanGEM/README.md`

mihai-sysbio:

This PR resolves #222. It also includes other styling fixes in `protocol.m`.  Also, it implements fixes for making web request from Matlab by always having a timeout of 30 seconds (larger than the default 5s).

_Note: the PR is in draft more, as it seems there might be more formatting to be done here_


edkerk (approved):
Looks good to me.

edkerk:
The reason for different styling in [663d1f9](https://github.com/SysBioChalmers/GECKO/pull/247/commits/663d1f99a2ee33eac9f3091fd68a26a7ed49a48e) is to make it easy to quickly which step of the `runDLKcat` function is running. Between those _headings_ is then output that comes from running the sytem terminal.

mihai-sysbio:
> The reason for different styling in [663d1f9](https://github.com/SysBioChalmers/GECKO/pull/247/commits/663d1f99a2ee33eac9f3091fd68a26a7ed49a48e) is to make it easy to quickly which step of the `runDLKcat` function is running. Between those _headings_ is then output that comes from running the sytem terminal.

The Docker setup doesn't benefit from the `===`.

Regarding the Python setup, I can see the argument. Still, it's sending a diy/inconsistency vibe. If this output highlight is needed, it would look better if say each step in the protocol would use such a separator. But then I guess what you're after is separating the outputs _within_ DLKcat between themselves. Thinking from the global GECKO perspective, is that level of detail needed?

edkerk:
It just makes troubleshooting `runDLKcat` easier. Otherwise there are very few functions that print output to command window, and if they do, they do not need any separator. And if there would be such a need, it would be okay to use this style?

mihai-sysbio:
I'm basing my thinking largely on the [separation of concerns](https://en.wikipedia.org/wiki/Separation_of_concerns).

From the perspective of GECKO, troubleshooting `runDLKcat` is important. However, my expectation is that it would be to a large extent the consequence of not using the Docker approach. (Side-note, I believe I have some improvements on the Docker setup).

If instead we are talking about troubleshooting `DLKcat` itself, while being a valid concern, I would say it's outside of the scope of GECKO, so it should be resolved elsewhere.

@edkerk I believe you have the veto rights on all commits. Let me know what you decide when this PR becomes ready for review.

edkerk:
It's not about trouble shooting `DLKcat`, just the GECKO function `runDLkcat`. Besides the disp messages, is this PR otherwise ready for merging?

mihai-sysbio:
> Besides the disp messages, is this PR otherwise ready for merging?

I would have preferred to keep it in draft mode for more updates, but it makes more sense to merge it sooner rather than later. Any other issues/improvements will be dealt with separately.

edkerk:
Right, because I also have parallel changes that would be good to be merged together.

PR #251, opened 2023-02-28
feat: improve dlkcat docker

mihai-sysbio:

The changes in this PR drastically reduce the size of the Docker image containing DLKcat to about 1/3, and also the build time. Moreover, it openly shares the code to create the Docker image so that the image can be build by individual users. This code has been extracted from the branch `dlkcatPackage` and placed under `src/dlkcat-docker`, with the unnecessary files being removed.

~Therefore, upon merging this branch, the `dlkcatPackage` may be deleted.~ see [this comment](https://github.com/SysBioChalmers/GECKO/pull/251#issuecomment-1449921689)

Finally, the resulting Docker image has been published as a package on the GH Container Registry and linked to this repository.

_Note: ~This PR is based on #247 so it requires that to be merged to `main` first.~ done_


edkerk:
Do we need to somehow maintain/update the Docker image on this GitHub registry?

mihai-sysbio:
> Do we need to somehow maintain/update the Docker image on this GitHub registry?

Not more than with any Docker image. The image will stay like this indefinitely, but of course if it proves to be buggy it should be updated. The current image is tagged as `0.1`. While it could be overwritten, it think it would be better to just release newer versions. The image is linked from the homepage of the repository, and also [under the SysBioChalmers org](https://github.com/orgs/SysBioChalmers/packages?repo_name=GECKO).

Yu-sysbio:
I did not review the code but just tested by running `runDLKcatFromDocker()`, I got below:

Running DLKcat prediction, this may take many minutes
Unable to find image 'ghcr.io/sysbiochalmers/gecko3-dlkcat:0.1' locally
docker: Error response from daemon: Head "https://ghcr.io/v2/sysbiochalmers/gecko3-dlkcat/manifests/0.1": unauthorized.
See 'docker run --help'.
Error using runDLKcatFromDocker (line 45)
DLKcat encountered an error or it did not create any output file.

edkerk:
The image was set _private_. It should be public can, can you try again?

Yu-sysbio:
> The image was set _private_. It should be public now, can you try again?

Now it works on my computer. Well done!

Yu-sysbio:
Just confirm: the `GECKO/dlkcat` folder is not generated by using this improved dlkcat docker, correct?

And running `runDLKcat()` will still generate the `GECKO/dlkcat` folder?

edkerk:
DLKcat is now under /src, and `runDLKcat` needs to be refactored to also use that location.

mihai-sysbio:
> The image was set _private_. It should be public now, can you try again?

Not sure why, I want to remember having set it to _public_. Thanks for fixing this @edkerk.

ae-tafur (approved):
Nice, I tested an works on my computer,

mihai-sysbio:
> DLKcat is now under /src, and `runDLKcat` needs to be refactored to also use that location.

Actually, I was wondering about that. While it can be done, I'm not totally sold on the advantages of maintaining the (potentially harder to debug) Python setup.

Having the Python setup is definitely more convenient when frequently modifying `DLKcat.py`. However, my expectation is that most users will not do that (maybe even none?). However, with the source code for the Docker image in the `main` branch, anyone can modify `DLKcat.py` or any of the input files and run `docker build .` inside the `src/dlkcat-docker` folder. The resulting image would then have to replace `ghcr.io/sysbiochalmers/gecko3-dlkcat:0.1` in a single place in the `runDLKcatFromDocker()` code.

edkerk:
Is there a reason why the torch requirement description is changed? It is now hard-coded for  `linux_x86_64`:

Earlier: `torch>=1.8.0`
Now: `torch@https://download.pytorch.org/whl/cpu/torch-1.9.1%2Bcpu-cp39-cp39-linux_x86_64.whl`

Also, the `/src/dlkcat-docker` folder may be renamed to `/src/dlkcat`, but will this somehow brake `runDLKcatFromDocker` or the Docker? I don't see anywhere any reference to that folder name, but I guess I don't fully grasp how the Docker is generated (it appears [here](https://github.com/SysBioChalmers/GECKO/pkgs/container/gecko3-dlkcat), but how/why?).

mihai-sysbio:
> Is there a reason why the torch requirement description is changed? It is now hard-coded for `linux_x86_64`:
> 
> Earlier: `torch>=1.8.0` Now: `torch@https://download.pytorch.org/whl/cpu/torch-1.9.1%2Bcpu-cp39-cp39-linux_x86_64.whl`

This hard-coding avoids the installation of the `cuda` variant of `torch`, which inflated the Docker image by about 3 gb, and which also increased the build time because of having to download a lot more files. On my 4yo MacBook Pro the running time was about 15 min for the prediction using the CPU, so the use of GPU is not critical. Moreover, in Docker, the use of GPUs has been more tricky to set up, so there were clear advantages to restrict `torch` to the CPU version.

> Also, the `/src/dlkcat-docker` folder may be renamed to `/src/dlkcat`, but will this somehow brake `runDLKcatFromDocker` or the Docker? I don't see anywhere any reference to that folder name, but I guess I don't fully grasp how the Docker is generated (it appears [here](https://github.com/SysBioChalmers/GECKO/pkgs/container/gecko3-dlkcat), but how/why?).

This needs a more elaborate answer.

The contents of the git-ignored `/dlkcat` directory is identical to and was the initial source for `src/dlkcat-docker`. However, to not break the Python setup, these folders have been left independent. One could update the Python setup to use the code in `src/dlkcat-docker`, which would then be okay to be renamed to `src/dlkcat`, but that requires more work and was not envisioned to be part of this PR. This made me realise that the branch `dlkcatPackage` needs to kept until refactoring is done.

To generate the Docker image, one needs to follow the [instructions from GitHub](https://docs.github.com/en/packages/learn-github-packages/connecting-a-repository-to-a-package#connecting-a-repository-to-a-container-image-using-the-command-line), briefly presented here: 
- revise the version label in the `Dockerfile`
- run `docker build --tag gecko3-dlkcat .` inside `src/dlkcat-docker`
- create a GitHub personal access token (PAT)
- use `docker push` to publish the image to the GH Container registry
- make sure the image shows up as public on GitHub

mihai-sysbio:
@edkerk fyi even through @ae-tafur and @Yu-sysbio have tested this, I will wait for your formal approval before merging.

edkerk:
I'll test it ASAP

edkerk:
Now tested, it works smoothly. I'm ready to let go of `runDLKcat` :) and replace it with `runDLKcatFromDocker` as default.

However, I think it is good that we do allow the user the option to run the GECKO-DLKcat package outside of Docker. Now that DLkcat code is distributed via the repo (on Git LFS), it makes sense to make this also suitable for "native" installation.

Proposed changes:
- [ ] Rename `src/dlkcat-docker` to `src/dlkcat`
- [ ] Add `requirements_noDocker.txt` with the arch-agnostic `torch>=1.8.0`. (Alternative is to rename the current `requirements.txt` to `requirementsDocker.txt`, and use `requirements.txt` for the arch-agnostic variant. Either option is fine by me, but should be documented in the protocol).
- [ ] Rename `runDLKcatFromDocker` to `runDLKcat`
- [ ] Delete `dlkcatPackage` branch
- [ ] Update in protocol text and protocol.m (when all of above are done).

mihai-sysbio:
Alright!

> * [ ]  Rename `src/dlkcat-docker` to `src/dlkcat`

Nothing against it - I'm just thinking this folder ought to be named in a way that makes it clear it is not the original DLKcat. Perhaps also have a README inside that folder to clarify this?

> * Add `requirements_noDocker.txt` with the arch-agnostic `torch>=1.8.0`. (Alternative is to rename the current `requirements.txt` to `requirementsDocker.txt`, and use `requirements.txt` for the arch-agnostic variant. Either option is fine by me, but should be documented in the protocol).

I think this can also be solved by overloading the `torch` requirement in the installation command in the `Dockerfile`.

edkerk:
> Alright!
> 
> > * [ ]   Rename `src/dlkcat-docker` to `src/dlkcat`
> 
> Nothing against it - I'm just thinking this folder ought to be named in a way that makes it clear it is not the original DLKcat. Perhaps also have a README inside that folder to clarify this?

Fair point, but then `dlkcat-gecko` would be more accurate, as it is definitely GECKO specific, but not necessarily docker specific.

mihai-sysbio:
> Fair point, but then `dlkcat-gecko` would be more accurate, as it is definitely GECKO specific, but not necessarily docker specific.

Would you then suggest to also remove the name of the Docker image to `dlkcat-gecko:0.1`?

edkerk:
Perhaps that's best? It's up to you, you're obviously more fluent in Docker images et al.

mihai-sysbio:
Thank you all for testing and reviewing.

The scope of the PR has been expanded a bit as listed in [this comment above](https://github.com/SysBioChalmers/GECKO/pull/251#issuecomment-1450257343). Briefly, the new changes have to do with relying solely on the Docker approach for running DLKcat, while keeping the `requirements.txt` generic.

With a new (in name only) Docker image pushed to GHCR and the old one deleted, I will ask again for reviews.

mihai-sysbio:
_Note:_ before merging, there are two outstanding items from [this comment](https://github.com/SysBioChalmers/GECKO/pull/251#issuecomment-1450257343):

- [ ] Delete dlkcatPackage branch
- [ ] Update in protocol text and protocol.m (when all of above are done)

Yu-sysbio:
This (i.e., relying solely on the Docker approach for running DLKcat) works well on my computer. Well done!

edkerk (approved):

edkerk:
Failing test is known issue of https://github.com/SysBioChalmers/RAVEN/issues/469

Issue #249, opened 2023-02-28
bug for applyComplexData for rat-GEM


feiranl:
### Description of the bug:

As reported by external user @yanghong-bio, the search for complex data of rat is not working as we assumed. 

#### Current behavior:

![IMG_0502](https://user-images.githubusercontent.com/32157802/221905965-af224798-0a54-444e-84ef-ae20c415526c.JPG)

#### Reproducing these results:
[ecrat.mat.zip](https://github.com/SysBioChalmers/GECKO/files/10852246/ecrat.mat.zip)
```matlab
% If you worked on matlab, paste your code here
```

```python
# If you worked on python, paste your code here
```

ae-tafur:
Can you share the .json file dowloaded with the `getComplexData` ?

yanghong-bio:
[ComplexPortal.json.zip](https://github.com/SysBioChalmers/GECKO/files/10853389/ComplexPortal.json.zip)

PR #253, opened 2023-03-01
Fix: error calculating the match percentage

ae-tafur:

Solves #249 

Tested with data provided in #249, comments:

https://github.com/SysBioChalmers/GECKO/issues/249#issue-1603338884

https://github.com/SysBioChalmers/GECKO/issues/249#issuecomment-1448623950


edkerk (approved):
I haven't tested it, but if it solves the reported issue and passes the tests here it's fine for me!

PR #256, opened 2023-03-01
fix: check for customkcats file resolves #255

### LINKED ISSUE #255: no error reported when customKcats file is missing
### Description of the bug:

When trying to apply custom kcats in the absence of the custom kcats file, the pipeline throws a confusing error.

#### Expected behavior:

The pipeline is expected to check if the required file exists before proceeding.

mihai-sysbio:

This PR resolves #255 and slightly improves the error messages.

### DIFF
--- a/src/geckomat/change_model/applyCustomKcats.m
+++ b/src/geckomat/change_model/applyCustomKcats.m
@@ -49,10 +49,15 @@ if nargin < 3 || isempty(modelAdapter)
         error('Either send in a modelAdapter or set the default model adapter in the ModelAdapterManager.')
     end
 end
+
 params = modelAdapter.params;
 
 if nargin<2 || isempty(customKcats)
     customKcats = fullfile(params.path,'data','customKcats.tsv');
+end
+
+if ~exist(customKcats, 'file')
+   error(['Custom kcats cannot be applied because of missing expected file: ' customKcats]);
 end
 
 if isfile(customKcats)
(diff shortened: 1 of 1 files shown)


src/geckomat/change_model/applyCustomKcats.m
```diff
@@ -49,12 +49,17 @@
...
 end
+
 params = modelAdapter.params;
 
 if nargin<2 || isempty(customKcats)
     customKcats = fullfile(params.path,'data','customKcats.tsv');
 end
 
+if ~exist(customKcats, 'file')
+   error(['Custom kcats cannot be applied because of missing expected file: ' customKcats]);
```
edkerk, line 60: Missing file should not throw an error, as this is not an essential function.

```suggestion
   disp(['No custom kcats are applied due to missing customKcats file: ' customKcats]);
   rxnUpdated = [];
   notMatch = [];
   return
```
A wrongly formatted file is more serious, so the other error on line 76 should stay.

mihai-sysbio: If I got this right, the whole point of `applyingCustomKcats` is to read these custom kcats from a file into the model structure. If there is no file, then there is no point of calling the function.

For example, if I create a file name with a typo, what should the pipeline do:
1. Warn me that no kcats have been applied, which assumes I carefully read the entire pipeline output.
2. Crash, thus making it obvious for me that this command is not doing what is expected of it.

edkerk: I was thinking of a pipeline, where the same list of functions are applied to multiple models. Sometimes the model has a `customKcats.tsv`, sometimes it doesn't. But I suppose in that pipeline you could first include a check if the file exists, before running the function.
Most often the functions are not run within such a pipeline though, and then indeed an error message would be warranted. I'll accept!

edkerk (approved):

Issue #221, opened 2023-02-21
feat: improve main README


mihai-sysbio:
### Description of the new feature:

The main README file should be updated:
- [x] ~expand the clarification of the differences between `geckomat` and `geckopy` (especially that these are likely to have diverged even more)~
- [x] add link to tutorials from Readme
- [x] ~update `Contributors` section~
- [x] remove `Last update` date
- [x] either remove TravisCI label, or create a new issue to fix the CI pipeline
- [x] add Zenodo integration and add a Zenodo label
- [x] point to Protocol.m, users may not know the protocol is stored in `tutorials`
- [x] document the branch structure as defined in #227 in the Development Guidelines, and link to this from the README 
- [x] document the deletion of `geckopy` in #259 
- [x] replace `readthedocs` with `gh-pages`

edkerk:
Good point, README will be overhauled. It should also mention GECKO2 as well as GECKO3, with GECKO2 on a separate branch.

Yu-sysbio:
In practice, we should also friendly guide users, e.g. reviewers, who might be not familiar with GitHub, to GECKO3 rather than the previous versions.

mihai-sysbio:
While I am in support of that @Yu-sysbio, I still wonder how do deal with the `geckomat` vs `geckopy` conundrum.

edkerk:
Leave out `geckopy` until it is refactored. The current protocol is focusing on MATLAB.

mihai-sysbio:
Personally, I am fine with whatever decision regarding `geckopy`, but wanted to highlight that GECKO3 is only superseeding the Matlab versions of GECKO. Therefore, setting up a clear README is less straightforward that I'd wish.

feiranl:
Also point to Protocol.m, users may not know the protocol is stored in tutorials.

mihai-sysbio:
> Also point to Protocol.m, users may not know the protocol is stored in tutorials.

Great point @feiranl, I've added this to the issue checklist.

mihai-sysbio:
With #326 merged, this issue is fully resolved and will therefore be closed.

PR #259, opened 2023-03-01
chore: remove geckopy

edkerk:
Remove geckopy functionality that would be incompatible with the GECKO3 formulation. Even the S-matrix (MW/kcat instead of 1/kcat) is different, so the code is also not suitable for "only" simulating.

In the slightly longer term, it would be very benificial to rework this code, starting with the now streamlined GECKO3 base.


mihai-sysbio:
There is now a to-do to document this change in the README as part of #221.

PR #273, opened 2023-03-06
feat: stop if not limiting proteins were found

ae-tafur:

- Stop if all the coefficients are equal to zero during protein flexibilization.

- Only add coefficient if increasing the protein concentration increase growth rate


edkerk:
If no more proteins are limiting, the function will now also test if the protein pool exchange is limiting. Either way, some suggestions are given on how to proceed.

edkerk:
@ae-tafur Could you test this with your code?

ae-tafur:
> @ae-tafur Could you test this with your model + data?

Wow, this looks like a nice solution. Here are my results. 

I have three yeast strain. Which grows up to 0.41, 0.127, and 0.17. 

Under version 3.0.0 the maximum growth rate after proteomics integration 0.11. Updating and verifying some custom kcat values I was able to reach, 0.204, 0.127, and 0.13, respectively. 

With the last changes you commit in this PR:

First, 

> Strain 1: No (more) limiting proteins have been found. Attempt to increase protein pool exchange.
Changing the lower bound of protein pool exchange from -7.7632 to -8.7869 enabled a
growth rate of 0.2219. It can be helpful to set the lower bound of protein
pool exchange to -1000 before running flexibilizeProtConcs.

> Strain 2: It was working before. 

> Strain 3: No (more) limiting proteins have been found. Attempt to increase protein pool exchange.
Changing the lower bound of protein pool exchange from -9.3295 to -9.6186 enabled a
growth rate of 0.14091. It can be helpful to set the lower bound of protein
pool exchange to -1000 before running flexibilizeProtConcs.

Then, I follow the suggestions, set the lower bound of prot_pool_exchange, getting:

> Strain 1: Grow: 0.419

> Strain 2: Grow: 0.127 (it was working)

> Strain 3. Grow: 0.171

So, now I can reach the experimental conditions. That's magnificent!

> Note 1. I remove all my proteins with values lower than 0.05 ug/gDW

> Note 2. I update prot_pool_exchange. It pass from -125 to -7.76321667715018, -21.1878469176347, and -9.32947537640246, respectively

edkerk:
Sweet. Now I'm just wondering how we can make sure the user remains aware of the various flexibilizations that are done. The prot_pool change is also reported in the `flexData` structure, how to present this further will be another Issue.

edkerk (approved):

ae-tafur:
> Sweet. Now I'm just wondering how we can make sure the user remains aware of the various flexibilizations that are done. The prot_pool change is also reported in the `flexData` structure, how to present this further will be another Issue.

But, `flexibilizeProtConcs` return the changes made, Maybe it can just return that and maybe as an option save the results, in a tsv file in output folder, is user want to save it?

edkerk:
Yes, but less make a separate Issue and PR, because similar functionality would be good for various functions, and it would be good to implement this in a coherent way.

ae-tafur:
> Yes, but less make a separate Issue and PR, because similar functionality would be good for various functions, and it would be good to implement this in a coherent way.

Yeap, you're right.

Issue #279, opened 2023-03-14
Minor enhancement in protocol.m


Yu-sysbio:
1. https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/protocol.m#L72

    If one skips this line, then running Line 73 leads to the error:

    > Unrecognized function or variable 'complexInfo'.

2. https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/protocol.m#L87

    It is nice to display "The following reactions have invalid/incomplete EC numbers, correct them in model.eccodes and rerun getECfromGEM:", but it would be better to also suggest the correct format of the EC numbers maybe with an example like "'1.1.2.4;1.1.99.-'".

    And I just noticed that if there are any invalid/incomplete EC numbers, this function will populate `model.ec.eccodes` with all {''}. Even for the reactions with valid EC numbers, they will also be populated with {''}. Is this correct? If the user decides to not correct the invalid/incomplete EC numbers, then the ecModel will have no EC number populated in this step. My expectation would be that this function can at least populate EC numbers for the reactions with valid info.

3. https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/protocol.m#L97

    Even though there is a `smilesDB.tsv` file, this line queries PubChem every time.

4. Line 119 and Line 132 are identical.

5. https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/protocol.m#L139

    Would it be possible to directly use `obj.params.f` instead of `f` here if the user skips Line 138?

Edit @edkerk:
Actions:
- [x] 1. Update protocol.m complexInfo mention.
- [x] 2. Refactor getECfromGEM to report output of improper ECs
- [x] 3. Refactor findMetSmiles to also fill `smilesDB.tsv` with empty entries for failed PubChem queries
- [x] 4. Document rational for repeated `applyKcatConstraints` runs.
- [x] 5. Document that either `calculateFfactor` or the default 0.5 can be used.

edkerk:
[...]

PR #284, opened 2023-03-15
fix: minor enhancements in protocol

edkerk:
- solves #279 


src/geckomat/change_model/findMetSmiles.m
```diff
@@ -66,32 +66,35 @@
...
             error('Cannot reach PubChem. Check your internet connection and try again.')
         end
+        end
+        % Append one line each time, in case internet connection is lost
+        % halfway. Open & close file each time to avoid leaving the file
+        % open when breaking the function.
+        out = [uniqueNames(i), uniqueSmiles(i)];
+        fID = fopen(smilesDBfile,'a');
+        fprintf(fID,'%s\t%s\n',out{:});
+        fclose(fID);
```
mihai-sysbio, line 95: I remember this code struck me the first time I saw it. To my understanding, the aim here is to handle situations when the connection drops while this function is running. Why does the input need to be written to file every line? how about using a temporary data structure to hold all the entries, and do the writing in one go? It would save some time having to open and close a file several thousand times.

Also, while it looks like the core was built to handle mid-way interruptions, neither the documentation nor the output make this clear. It would be beneficial for the users to know that X entries were retrieved from the file before the querying starts.

edkerk: Good suggestions, but I don't have the resource to refactor this at the moment.

protocol.m
```diff
@@ -126,17 +134,32 @@
...
+% applyKcatConstraints is rerun as explained in step 11.
 ecModel = applyKcatConstraints(ecModel);
 
 % STEP 14 Set upper bound of protein pool
-% Calculate f-factor (how much proteins are enzymes), based on paxDB.tsv
-% that is provided for the model. Other proteomics data can also be used,
-% while an alternative approximation of 0.5 is typically quite realistic.
-f = calculateFfactor(ecModel);
-ecModel = setProtPoolSize(ecModel,[],f);
+% The protein pool exchange is constraint by the total protein content
```
Yu-sysbio, line 143: Line 143 `constraint` to `constrained`

src/geckomat/get_enzyme_data/getECfromGEM.m
```diff
@@ -1,6 +1,9 @@
-function model = getECfromGEM(model, ecRxns)
+function [model, invalidEC, invalidECpos] = getECfromGEM(model, ecRxns)
 % getECfromGEM
-%   Use the model.eccodes to populates the model.ec.eccodes field.
+%   Use the model.eccodes to populates the model.ec.eccodes field. EC
+%   numbers that are not formatted as four numbers separated by periods,
+%   possibly with trailing wildcards. Examples: 1.2.3.4 or 1.2.3.- while
+%   invalid EC numbers are 1.2.3 or 1_2_3_4.
```
Yu-sysbio, line 6: And also give an example: 1.2.3.4;1.2.3.5 is valid while 1.2.3.4|1.2.3.5 is invalid.

Yu-sysbio (commented):
Looks good!

protocol.m
edkerk:
```suggestion
% The protein pool exchange is constrained by the total protein content
```

src/geckomat/get_enzyme_data/getECfromGEM.m
edkerk:
```suggestion
%   invalid EC numbers are 1.2.3 or 1_2_3_4. Multiple EC numbers are
%   separated by ; for instance 1.2.3.4;1.2.3.5 not 1.2.3.4|1.2.3.5.
```

Yu-sysbio (commented):
Looks good!

Issue #287, opened 2023-03-20
Error in minimizing protein pool usage


Yu-sysbio:
https://github.com/SysBioChalmers/GECKO/blob/70b93d33aa5aefddeef2957c355046534a6b1ee6/protocol.m#L237

Should it be 1? The flux of prot_pool_exchange is negative.

ae-tafur:
Yes, now flux is negative it must be positive

edkerk:
Solved in 05df8105e61da016fb4d224834c7f1be0e914f97

Yu-sysbio:
> Solved in [05df810](https://github.com/SysBioChalmers/GECKO/commit/05df8105e61da016fb4d224834c7f1be0e914f97)

Actually there are a few lines containing `-1` that should be corrected in the `protocol.m` file.
And let's keep in mind correcting this in the protocol manuscript.

Issue #286, opened 2023-03-20
ec.source in getStandardKcat


Yu-sysbio:
`getStandardKcat` allows to assign kcats for reactions without any associated genes, and then mark in `.ec.source` with `'standard'`.
It can also assign kcats for the reactions with genes but not assigned kcats by previous steps e.g. BRENDA and DLKcat, but the `.ec.source` is marked with empty, i.e. `''`. I am wondering if this should also be marked with `'standard'`.

Given that `'brenda'` and `'dlkcat'` mean the kcat source, does `'standard'` mean the kcat itself, or the enzyme, kcat and molecular weight as a whole?

ae-tafur:
I agree. 

> Given that `'brenda'` and` 'dlkcat'` mean the kcat source, does` 'standard'` mean the kcat itself, or the enzyme, kcat and molecular weight as a whole?

It means the kcat itself

Issue #280, opened 2023-03-15
getStandardKcat assigns protein costs to pseudo-reactions


Yu-sysbio:
This function [getStandardKcat](https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/src/geckomat/gather_kcats/getStandardKcat.m) assigned protein costs to many pseudo-reactions in my model, e.g.,Biomass, NGAM etc. 
What do you suggest to avoid this?

ae-tafur:
Your model in ecModel.rxnNames contains the word "pseudoreaction" ? 

If not, this can be the reason because why it happens.

A quick solution is to add this reactions in the adapter model method 'getSpontaneousReactions'

Yu-sysbio:
No, ecModel.rxnNames does not contain "pseudoreaction". 
While this is not a bug and will not cause any error, it would result in an ecModel with incorrect protein costs if the conventional GEM does not contain the pseudoreaction information.

It is very necessary to indicate somewhere in the protocol: before running getStandardKcat, please ensure to either include "pseudoreaction" in the rxnNames field of the starting model or add the reactions that you do not want to assign pseudo protein costs in the adapter to avoid this issue.

edkerk:
This will be explained in the protocol, plus a custom `data/nonEnzymeRxns.tsv` file can be specified, which `getStandardKcat` will try to read and if it exists, it will use it to exclude those reactions by their reaction identifier. So this is model-specific.

PR #288, opened 2023-03-20
feat: getStandardKcat use custom list of non-enzyme reactions to ignore

edkerk:
- feat:
  - `getStandardKcat` can use `data/pseudoRxns.tsv` (if it exists) to define custom non-enzyme reactions that should be ignored when assigning standard kcats. (solves #280)
  - `getStandardKcat` removes previous standard kcat assignments, so that the function can be run repeatedly on the same model.
- fix:
  - `getStandardKcat` also specifies `model.ec.source{i} = 'standard'` if a standard kcat is assigned to a zero-kcat reaction. (solves #286)
  - remaining corrections for prot_pool_exchange minimization (solves #287)


ae-tafur (approved):
Looks good!

protocol.m
```diff
@@ -130,8 +130,25 @@
 ecModel = getKcatAcrossIsoenzymes(ecModel);
 
 % STEP 12 Get standard kcat
-% Assign an enzyme cost to reactions without gene assocation (except
-% exchange, transport and pseudoreactions)
+% Assign an enzyme cost to reactions without gene assocation. These
```
Yu-sysbio, line 133: Maybe change `an enzyme cost` to `a protein cost` to keep consistent with the manuscript.

protocol.m
```diff
@@ -130,8 +130,25 @@
...
+% The following reactions are exempted:
+% A Exchange reactions: exchanging a metabolite across the model boundary,
+%   not representing a real enzymatic reaction.
+% B Transport reactions: transporting a metabolite with the same name from
+%   one compartment to another. Real transport reactions should already be
+%   annotated with grRules, so that the remaining non-annotated reactions
+%   are mostly representing diffusion or pseudotransport processes such as
+%   vesicles moving from ER to Golgi. While proteins are involved in such
+%   processes, they are not catalyzed by enzymes.
+% C Pseudoreactions: any other reaction that should be considered to be
```
Yu-sysbio, line 144: `should be considered` -> `should not be considered` ?

src/geckomat/gather_kcats/getStandardKcat.m
```diff
@@ -126,69 +133,88 @@
...
 
+% Get custom list of reaction IDs to ignore, if existing. First column
+% contains reaction IDs, second column contains reaction names for
+% reference only.
+if exist(fullfile(params.path,'data','pseudoRxns.tsv'),'file')
+    fID        = fopen(fullfile(params.path,'data','pseudoRxns.tsv'));
+    fileData   = textscan(fID,'%s %s','delimiter','\t');
+    fclose(fID);
+    customRxns = fileData{1};
+    customRxns = find(strcmp(customRxns,model.rxns));
```
Yu-sysbio, line 144: Error using `strcmp`
Inputs must be the same size or either one can be a scalar.

Yu-sysbio (commented):
Looks good!

protocol.m
edkerk:
```suggestion
% Assign a protein cost to reactions without gene assocation. These
```

src/geckomat/gather_kcats/getStandardKcat.m
edkerk:
```suggestion
    customRxns = find(ismember(model.rxns,customRxns));
```

Issue #277, opened 2023-03-09
Bug: Looks for the taxonomically closest organism for matching the Kcats


HossFir:
Description: There was an issue in fuzzyKcatMatching.m when it searches BRENDA for the closest relative organism when the Kcat is unavailable.

The type of error was:
The logical indices contain a true value outside of the array bounds.

I commented the line 493 to skip the issue.

https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/src/geckomat/gather_kcats/fuzzyKcatMatching.m#L493
https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/src/geckomat/gather_kcats/fuzzyKcatMatching.m#L357
https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/src/geckomat/gather_kcats/fuzzyKcatMatching.m#L316
https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/src/geckomat/gather_kcats/fuzzyKcatMatching.m#L214
https://github.com/SysBioChalmers/GECKO/blob/ea9e2ef56673f50b764a183fbe1a6d746f64cff3/src/geckomat/gather_kcats/fuzzyKcatMatching.m#L173

edkerk:
Would you be able to share the ecModel so that we can try to replicate the problem? This part of the code has been unchanged since GECKO2, so it _should_ not suddenly have a bug.

HossFir:
Sure! This is the ecModel for my organism:

[eciHK_fv.zip](https://github.com/SysBioChalmers/GECKO/files/11018751/eciHK_fv.zip)

edkerk:
I cannot replicate the error. What did you specify as `obj.params.org_name` in the ModelAdapter? And what MATLAB version are you running?

HossFir:
For GECKO 2, I followed the protocol given and downloaded the Prot_abundances from PaxDb and also retrieve Kcat values from BRENDA.
Using GECKO 3, I just set all these automatically in the ModelAdapter.m.

edkerk:
> obj.params.org_name = 'synechocystis sp. PCC 6803'

[...]

PR #289, opened 2023-03-20
fix: if organism is in KEGG multiple times

edkerk:
- fix:
  - `fuzzyKcatMatching` solves #277 when KEGG has multiple entries with the same organism name.  


ae-tafur (approved):

ae-tafur:
Look good !

mihai-sysbio:
The implementation of this fix can be very tricky to debug later on. If there is no better solution than choosing the first match, please also consider updating the documentation of this function.

edkerk:
See 886a15ec047cfdf32ed3ae24e50803f9ba2b97db and #295.

PR #295, opened 2023-03-21
fix: minor issues raised

edkerk:
### Main improvements in this PR:
- fix: minor issues raised in #288, #289

**I hereby confirm that I have:**

- [x] Selected `develop` as a target branch (top left drop-down menu)

PR #299, opened 2023-04-02
fix: getStandardKcat with multiple subSystems

edkerk:
- fix:
  - `getStandardKcat` handling models that have multiple subSystems assigned to the same reaction.

### DIFF
--- a/src/geckomat/gather_kcats/getStandardKcat.m
+++ b/src/geckomat/gather_kcats/getStandardKcat.m
@@ -4,7 +4,12 @@ function [model, rxnsMissingGPR, standardMW, standardKcat, rxnsNoKcat] = getStan
 %   used to apply enzyme constraints to reactions without any associated genes.
 %   This is done by adding those reactions to model.ec, assign a "standard"
 %   pseudoenzyme with the standard MW (median of all proteins in the organism)
-%   and standard kcat (media from all kcat, or subsystem specific kcat).
+%   and standard kcat (median from all kcat, or subsystem specific kcat).
+%
+%   A reaction is assigned a subSystem specific kcat values if the model
+%   has a subSystems field and the reaction is annotated with a subSystem.
+%   Only the first subSystem will be considered if multiple are annotated
+%   to the same reaction.
 %
 %   Exchange, transport and pseudoreactions are filtered out, plus any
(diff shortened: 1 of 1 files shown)


ae-tafur (approved):

mihai-sysbio:
The output of the action looks unexpected, but also doesn't seem to be caused by changes in this PR. Should it be investigated further (perhaps as a separate issue)?

> Model-specific DLKcat input stored at /home/m/actions-runner/_work/GECKO/GECKO/test/unit_tests/ecTestGEM/data/DLKcat_input_test.tsv
> Limit has been reached. Protein P5 seems to be problematic. Consider changing the kcat.

ae-tafur:
I think is ok, this is output from the functions tested. `runDLKcat` (not 100% sure), and `flexibilizeProtConcs.m` (when a limit is set)

ae-tafur:
> The output of the action looks unexpected, but also doesn't seem to be caused by changes in this PR. Should it be investigated further (perhaps as a separate issue)?
> 
> > Model-specific DLKcat input stored at /home/m/actions-runner/_work/GECKO/GECKO/test/unit_tests/ecTestGEM/data/DLKcat_input_test.tsv
> > Limit has been reached. Protein P5 seems to be problematic. Consider changing the kcat.

I think is ok, this is output from the functions tested. `runDLKcat` (not 100% sure), and `flexibilizeProtConcs` (when a limit is set)

mihai-sysbio:
Other recent PRs have a much more verbose output when running the test suite, e.g. https://github.com/SysBioChalmers/GECKO/pull/295#issuecomment-1477424625 . Perhaps I missed something, but I was expecting a similar output here.

ae-tafur:
> Other recent PRs have a much more verbose output when running the test suite, e.g. [#295 (comment)](https://github.com/SysBioChalmers/GECKO/pull/295#issuecomment-1477424625) . Perhaps I missed something, but I was expecting a similar output here.

Yeap, that was because a test was failing (check more [here](https://github.com/SysBioChalmers/RAVEN/pull/473)), but due to RAVEN reversibility assignment. This was fixed [here](https://github.com/SysBioChalmers/RAVEN/tree/v2.8.0)

Issue #229, opened 2023-02-22
chore: consistent names and terminology


edkerk:
With the intense development work over the last months, there is not always consistency on what kind of terminology is used. It is helpful to be consistent, and it is good to enforce this from the first GECKO3 release onwards.

### 🚨 **Suggestion: changes in function names will be applied in the protocol and here on GitHub on Sunday night 26 February.**

Note that the example function names indicated below are _after_ changing them to their new name.

| Term |  Meaning / use |  Example |
|  ----  |  ----  |  ----  |  
| ecModel |  abbreviation of enzyme-constrained model | |
| empty ecModel |  ecModel after `makeEcModel`, without kcat values | |
| modelAdapter |  a model-specific version of `ModelAdapter.m` | `ecYeastGEMAdapter.m` |
| `kcatList` |  structure that contains kcat information, can have multiple entries per reaction / enzyme / source  | output of `fuzzyMatching`, `loadDLKcatOutput()` |
| `load...()` | function that reads a datafile into MATLAB | `loadEcModel()` |
| `save...()` | function that writes a datafile from MATLAB | `saveEcModel()` |
| `gather...()` | function that downloads data or places API calls to an online resource. The output of this might be a data file, data structure in MATLAB, or directly in an `ecModel`  | `gatherMetSMILES()`
| `fill...()` | function that downloads data or places API calls to an online resource  | `fillKcat()`
| `apply...()` | function that modifies an ecModel with information that it already contains. For instance: `applyKcatConstraints` takes the information from `ecModel.ec.mw` and `ecModel.ec.kcat` to modify `ecModel.S`. | 

Not all function names will be covered by this scheme. For instance `fuzzyMatching` does not fit any of the categories. Or should we have `doFuzzyMatching`? Or specify `get...()` to indicate a function that applies some analysis on the model and gives an output that is outside the model (in this case a `kcatList`)?

This list is not complete yet, **feel free to add suggestions below.**

ae-tafur:
I thing `get...()` fits better . Then, following your example. 
| Term |  Meaning / use |  Example |
|  ----  |  ----  |  ----  |  
| `get...()` | gives an output that is outside the model | `fuzzyKcatMatching()` |
| `update...()` | function that update a value calculate previously  | `updatePoolSize()` |

PR #301, opened 2023-04-05
Bug: Fix compatibility running ecFSEOF with GECKO3 models

ae-tafur:

- Fix compatibility in ecFSEOF to GECKO3 models
- Change name function `simulateGrowth` to follow #229 
- Model adapter class was added to use `userData/model/output` as default folder to store results


src/geckomat/utilities/ecFSEOF/run_ecFSEOF.m
```diff
@@ -1,53 +1,72 @@
...
+% for a specified production target.
 %
-% Usage: run_ecFSEOF(ecModel,rxnTarget,cSource,alphaLims,Nsteps)
+% Input:
+%   ecModel         an ecModel in GECKO 3 format (with ecModel.ec structure).
+%   rxnTarget       rxn ID for the production target reaction, a exchange
+%                   reaction is recommended.
+%   cSource         rxn ID for the main carbon source uptake reaction.
+%   alphaLims       vector of Minimum and maximum biomass yield [gDW/mmol Csource]
+%                   for enforced objective limits (e.g. [0.5 1]). Max value: 1.
```
edkerk, line 13: I don't understand why the maximum value should be 1?

ae-tafur: Alpha is scaling factor for desired suboptimal growth (which fix the LB of biomass reaction), then a value higher than 1 will result in non feasible solutions. e.g if the maximum growth is 0.41 an scaling factor of 1.1 mean fix the LB to 0.451, a value that cannot be reached.

edkerk: Ok, but then the description is somewhat confusing. I thought that these values are gDW/mmol Csource, but they are actually scaling factors of the experimental growth rate.

ae-tafur: Yeap, you are right I will change it

src/geckomat/utilities/ecFSEOF/run_ecFSEOF.m
```diff
@@ -1,53 +1,72 @@
...
 %
-% Usage: run_ecFSEOF(ecModel,rxnTarget,cSource,alphaLims,Nsteps)
+% Input:
+%   ecModel         an ecModel in GECKO 3 format (with ecModel.ec structure).
+%   rxnTarget       rxn ID for the production target reaction, a exchange
+%                   reaction is recommended.
+%   cSource         rxn ID for the main carbon source uptake reaction.
+% 	alphaLims       vector of Minimum and maximum biomass scalling factors for
+%			        enforced objective limits (e.g. [0.5 1]). Max value: 1.
+%   Nsteps          number of steps for suboptimal objective in FSEOF.
```
edkerk, line 14:
```suggestion
%   cSource         rxn ID for the main carbon source uptake reaction.
%   alphaLims       vector of Minimum and maximum biomass scalling factors for
%                   enforced objective limits (e.g. [0.5 1]). Max value: 1.
%   Nsteps          number of steps for suboptimal objective in FSEOF.
```

edkerk (approved):

Issue #294, opened 2023-03-21
sensitivityTuning can introduce changes in undesired reactions


ae-tafur:
### Description of the new feature:

When running `sensitivityTuning`, it can make changes in reactions that results in undesired behavior of the model (compared with experimental data). A solution can be allow to define some reactions to be ignored in `sensitivityTuning`

**I hereby confirm that I have:**
- [ ] Done this analysis in the `main` branch of the repository
- [ ] Checked that a similar issue does not exist already

PR #306, opened 2023-04-24
Feat: ignore in tunning kcats user defined proteins

ae-tafur:

- User can define a list of protein which will not be considered in `sensitivityTuning`. 
- Solves #294 
- Tested with both full and light ecModels in yeast following the `protocol.m`


mihai-sysbio:
@ae-tafur ✅

edkerk (approved):

Issue #308, opened 2023-04-25
restructure tutorials


mihai-sysbio:
Based on team discussions, it it proposed to alter the `userData` folder and `protocol.m` as follows:

- [x] Rename `userData` to `tutorial`
- [ ] Mention the ecYeastGEM that is generated in GECKO3 explicitly as tutorial model, so the protocol constructs a tutorial ecModel which is based on S. cerevisiae. This means that e.g. various model-specific filenames would include _tutorial or something similar.
- [ ] Because of the above, it is very important to (repeatedly) mention that the model generated in protocol.m is a tutorial model. It will also be included in the filename, explicitly mentioned somewhere in the text.
- [ ] make sure there is a functional `protocol.m` for `tutorial_Human-GEM`

mihai-sysbio:
Added another to-do to provide a `protocol.m` for `tutorial_Human-GEM`.

edkerk:
The idea has been to have `protocol.m` match the manuscript text. That means, `protocol.m` generates a model for _S. cerevisiae_, and the user could have the MATLAB code open while reading through the paper. With that in mind, it is less convenient to have the detailed comments in blankExample instead.

Meanwhile, the idea of blankExample was to provide a suggested minimal configuration of folder structure and model adapter file. In that light it makes less sense to include a script with commands that refer to a specific yeast model.

So my suggestion is to leave out protocol from blankExample, and keep the detailed comments in the protocol in tutorial_yeast.

mihai-sysbio:
I'm thinking as a user, what would I do - where would I start from?
a) An existing `tutorial_`. The tutorial should "just work" as-is. Since probably the next step is to modify some parameters, I will then need to understand more about these so I will look at the documentation, which is the full `protocol.m` and the generic `adapter`. For me it's normal for documentation to sit elsewhere from the tutorial/example.
b) An empty canvas, the `blankExample`. Here I probably need the full protocol and generic adapter, so I can start coding. Probably the documentation can be elsewhere, but I will have go through it closely since I am not familiar with all the parameters and options. To me, it is an improved experience to have the documentation _in_ the blank example so that as a user I only need that folder to have a complete picture.

Now, here is what I think is potentially very tricky:
> The idea has been to have protocol.m match the manuscript text. 

and

> the user could have the MATLAB code open while reading through the paper

[...]

Issue #305, opened 2023-04-23
Bug detected


Champion-GuanWang:
#### Reproducing these results:

```matlab
% If you worked on matlab, paste your code here
```

#### System information

* Operating system: 
* MATLAB version: 
* GECKO version: 

**I hereby confirm that I have:**
- [ ] Tested my code with [all requirements](https://github.com/SysBioChalmers/GECKO) for running GECKO
- [ ] Done this analysis in the `main` branch of the repository
- [ ] Checked that a similar issue does not exist already

mihai-sysbio:
@Champion-GuanWang thanks for trying out the latest GECKO, for the bug report and for including the script. Could I ask you to provide a few more details on the exact versions of RAVEN, Matlab, and solver?

mihai-sysbio:
Also, do I understand correctly that you had no issues with `runDLKcat` in Docker?

Champion-GuanWang:
> Also, do I understand correctly that you had no issues with `runDLKcat` in Docker?
Yes, we had no issues with  'runDLKcat' in Docker. More details about Raven, Matlab and solver are as follows:
Matlab R2022a academic use; Raven 2.8.0; Set RAVEN solver gurobi 9.5.2. Thank you for debugging!

Champion-GuanWang:
> @Champion-GuanWang thanks for trying out the latest GECKO, for the bug report and for including the script. Could I ask you to provide a few more details on the exact versions of RAVEN, Matlab, and solver?
I am wondering there are any updates about the debugging? We are eager to have your reply. Thank you!

edkerk:
Your `solEC.x` is probably empty? The FBA that was run to generate `solEC` did not find a valid solution.

mihai-sysbio:
I was able to reproduce the issue on macOS, Matlab 2021b, RAVEN 2.8.0 and Gurobi 10, by running `protocol.m` in [this commit](ad2a5930dff80d3eefbbc3e101b5545f6739dc2f):
```
STEP 23 Compare fluxes from ecModel and starting model

solEC = 

  struct with fields:

       x: []
       f: []
    stat: -1
     msg: 'The problem is infeasible'
```

edkerk:
So the problem is the tutorial, not the function. 

We will be editing the tutorial during the next 1-2 weeks, to address this and some other aspects related to resubmission of the GECKO3 manuscript. This issue will be resolved as part of this process.

edkerk:
[...]

PR #310, opened 2023-04-27
chore: restructure tutorials

mihai-sysbio:
This PR aims to restructure the tutorials, resolving #308.

In short, it establishes self-standing, independent tutorials (with all subfolders like `userData` before, including the adapter). Each tutorial also has now its own protocol.
The way the adapters are handles has been slightly tweaked.

Expected to be restructured:
- [x] tutorial for yeast-GEM:
  - [x] stuck as step 23, reported independently [here](https://github.com/SysBioChalmers/GECKO/issues/305))
  - [x] should be the full `protocol.m`, with all comments
- [ ] tutorial for Human-GEM
  - [ ] using `geckoLight` variant
- [ ] tutorial for E.coli (@Yu-sysbio is working on this)
- [ ] ~blank example~
  - [ ] ~with generic `protocol.m` and `adapter`~


ae-tafur:
I think that adding the word tutorial as a prefix inside the tutorial folder may be redundant, isn't it ? or is there a specific reason to do this ?

mihai-sysbio:
> I think that adding the word tutorial as a prefix inside the `tutorial` folder may be redundant, isn't it ? or is there a specific reason to do this ?

Fair point, I also hesitated about this. My reasoning was that people will likely copy one of these folders, so when they do it will be clearer what is the "tutorial" what is what they modified. I might be wrong here, we should follow what is most intuitive for the users so it's good to have other opinions.

test/unit_tests/ecTestGEM/TestGEMAdapter.m
```diff
@@ -5,8 +5,8 @@
             
             %Directory where all model-specific files and scripts are kept.
             %Is assumed to follow the GECKO-defined folder structure. The
-            %code below refers to userData/ecYeastGEM in the GECKO path.
-            obj.params.path = fullfile(findGECKOroot,'test','unit_tests','ecTestGEM');
+            %code below refers to tutorials/tutorial_yeast-GEM in the GECKO path.
+            obj.params.path = fullfile(findGECKOroot,'test','unit_tests','ecTestGEM', 'TestGEMAdapter.m');
```
mihai-sysbio, line 9: The path was mistakenly altered here:
```suggestion
            obj.params.path = fullfile(findGECKOroot,'test','unit_tests','ecTestGEM');
```

mihai-sysbio: I'm confirming here the fix worked.

edkerk:
Recent changes:
- Remove `blankExample`, as it is replaced by `startGECKOproject` (see #312)
- Restore (and update) comments in ecYeastGEM `protocol.m` (as agreed upon in #308)
- Run through the ecYeastGEM protocol
- Simplified `ModelAdapterManager` commands (most relevant: `.getDefault` instead of `.getDefaultAdapterFromPath`)

Remaining:
- `ecHumanGEM` tutorial with GECKO light
- `ecEcoli` tutorial

I suggest to have these addressed in separate PRs, so that this PR with its significant refactoring of file locations can be merged with `develop`. I will open two Issues for the remaining tasks.

edkerk (approved):

PR #313, opened 2023-05-17
feat: currencyMets and pseudoreactions handling

edkerk:
Based on reviewer's comments:
- feat:
  - `makeEcModel` can use pseudoRxns.tsv to identify pseudoreactions
  - `writeDLKcatInput` only exclude currency metabolites if there are other reactants involved in the reaction


ae-tafur (approved):
Looks good

PR #316, opened 2023-05-31
fix: minor bugs identified

ae-tafur:

- Use logical vector for construct equation in ecFSEOF
- Change tolerance when fixing bound in ecFSEOF


edkerk (approved):
Looks good, I just added a check for `alphaLims` parameter, as it took me a while to understand that I did something wrong :)

PR #318, opened 2023-06-20
feat: online docs

### LINKED ISSUE #317: online documentation
### Description of the new feature:

While the code is documented, this documentation is not accessible online. The aim of this issue is to set up the same process as for RAVEN, where the inline documentation is put up online with a GH action.

mihai-sysbio:

This PR resolves #317 by adding the documentation with the necessary configuration such that it will become visible online. The link to the documentation is also included in the Readme.

_Note: the documentation will only be visible online once this code ends up on the `main` branch_


edkerk (approved):
Now including `updateGECKOdoc` function that prepares the HTML files. Uses the RAVEN provided m2html functionality, as RAVEN is a dependency anyway.

src/geckomat/utilities/updateGECKOdoc.m
```diff
@@ -0,0 +1,33 @@
+function updateGECKOdoc()
```
mihai-sysbio, line 1: Thanks for adding this @edkerk, I was too lazy to add it myself. Also, I kind of wanted to avoided adding it, since it is essentially the same as in RAVEN with the exception of the base directory and the folders to be avoided. A more reusable solution would be to generalise the function in RAVEN to take these as parameters, so then one could call a function to generate documentation for any "parent" folder that is desired. In my view, this PR achieves its purpose with or without this function.
This PR can proceed as-is, but if you agree @edkerk perhaps you could open an issue in RAVEN.

mihai-sysbio:
A deployment of the GH Pages site from the `develop` branch has validated that the documentation of the source-code is available online. Therefore, this issue is considered complete and will be closed.

src/geckomat/utilities/updateGECKOdoc.m
edkerk: I agree that a generic function would be good, but it's very low priority tbh.

PR #321, opened 2023-06-26
fix: full_ecModel tutorial ecFVA

edkerk:
- fix:
  - update ecFVA analysis in full_ecModel tutorial
  - `GECKOInstaller` do not add `doc` to MATLAB path
- feat:
  - `plotEcFVA` function that can be used to plot results from ecFVA
  - `saveEcModel` more intuitive input parameters (breaks backwards compatibility, corrected in protocols and manuscript)
- refactor:
  - `fuzzyKcatMatching` & `getEcFromGEM` speed improvement 


ae-tafur (approved):
There are some figures store as tiff file, is it not better to use pdf instead ?

mihai-sysbio:
> There are some figures store as tiff file, is it not better to use pdf instead ?

Either format is binary - how about storing all these binary files with git-lfs?

edkerk:
They're a few small files that will likely never be changed, git-lfs is not required.

PR #322, opened 2023-06-27
feat: light_ecModel tutorial getSubsetEcModel

edkerk:
- fix:
  - `getStandardKcat` avoid adding 'standard' pseudoprotein in _light_ ecModels
- feat:
  - `getSubsetEcModel` function that can contextualize ecModels based on (f)tINIT derived models
  - `light_ecModel` tutorial includes example of `getSubsetEcModel`, complete with an example contextualized HT-29 GEM as input.  

~After merging #321, this PR should be rebased to `develop`.~


mihai-sysbio:
I'm wondering if perhaps this would also be a good opportunity to address the minor issue around:
> writeDLKcatInput(ecModel);

Since there is a `DLKcat.tsv` in the `/data` folder, the code will trigger error that the file already exists. One option would be to specify all the parameters in the function, thus sending an `overwrite` argument. Another would be to rename the file, and write up a [README](https://github.com/SysBioChalmers/GECKO/blob/main/userData/yourModel/README.md) in each `tutorial` folder that would also cover `/data` and why the `DLKcat-backup.tsv` is provided.

edkerk:
With #321 merged, this PR can be reviewed.

> I'm wondering if perhaps this would also be a good opportunity to address the minor issue around:
> 
> > writeDLKcatInput(ecModel);
> 
> Since there is a `DLKcat.tsv` in the `/data` folder, the code will trigger error that the file already exists. One option would be to specify all the parameters in the function, thus sending an `overwrite` argument. Another would be to rename the file, and write up a [README](https://github.com/SysBioChalmers/GECKO/blob/main/userData/yourModel/README.md) in each `tutorial` folder that would also cover `/data` and why the `DLKcat-backup.tsv` is provided.

IMHO, I don't think it's a minor issue, I'm not too happy with either of these two options.
1: Overwriting the file with an empty DLKcat.tsv (without kcat values) forces the user to run DLKcat to continue with the tutorial, which is not ideal as running DLKcat can take long, requires Docker etc. Of course the user can just restore the file, or re-download the DLKcat.tsv, but this feels sub-optimal.
2: Having a differently named file makes the tutorials sub-folder not present a "normal" situation, with alternatively named files. The user might end up with two copies (DLKcat.tsv and DLKcat-backup.tsv), which is confusing.

A third option could be to command-out the writeDLKcatInput command in the protocol.m, and explain why this is done.

ae-tafur:
> A third option could be to command-out the writeDLKcatInput command in the protocol.m, and explain why this is done.

I agree with the third option.

src/geckomat/gather_kcats/getStandardKcat.m
```diff
@@ -150,14 +150,13 @@
...
 if ~any(strcmp(model.mets,'prot_standard'))
-    % Add a new gene to be consistent with ec field named standard
-    proteinStdGenes.genes = 'standard';
-    if isfield(model,'geneShortNames')
-        proteinStdGenes.geneShortNames = 'std';
-    end
-    model = addGenesRaven(model, proteinStdGenes);
-
     if ~model.ec.geckoLight
+        % Add a new gene to be consistent with ec field named standard
```
ae-tafur, line 154: Removing this will induce a difference in data between `model.ec.genes` and `model.genes`

ae-tafur (approved):

edkerk:
> > A third option could be to command-out the writeDLKcatInput command in the protocol.m, and explain why this is done.
> 
> I agree with the third option, but this means that `runDLKcat` will not work, right ? It needs the correct rxns id

I think you could run

> > A third option could be to command-out the writeDLKcatInput command in the protocol.m, and explain why this is done.
> 
> I agree with the third option, but this means that `runDLKcat` will not work, right ? It needs the correct rxns id

If the user wants to run `runDLKcat`, it will still be able to run either on the existing `DLKcat.tsv` file that already has kcat values (it will just overwrite those with new values, will double-check this right now), or with a new `DLKcat.tsv` if the user has uncommented `writeDLKcatInput`.

src/geckomat/gather_kcats/getStandardKcat.m
edkerk: You're right, might be best to revert this change. It caused problems with I/O a model (as the gene will not be in any grRule (and it shouldn't!)), but this will actually be solved here: https://github.com/SysBioChalmers/RAVEN/pull/494.

edkerk:
Will re-run the whole light_ecModel tutorial and save the final model after merging other PRs, as they will also have some minor impact.

PR #325, opened 2023-06-30
feat: parameters format, phylDist.mat file and yeast-GEM.yml

edkerk:
- rename KEGG, UniProt and Complex Portal related parameters from e.g. `obj.params.keggID` to `obj.params.kegg.ID`, to better group them together
- for Complex Portal, use taxonomic ID instead of species name, as the latter is hard to get exactly in the right format
- remove unused parameters from model adapters, remnants of GECKO1&2
- use RAVEN's `phylDist.mat` for `fuzzyKcatMatching`, which is more uptodate and avoids duplicate distribution of the same file (requires https://github.com/SysBioChalmers/RAVEN/pull/498, soon to be added to RAVEN 2.8.3 release)
- provide `yeast-GEM` in `yml` format in tutorial, to save space, but also make it less somewhat less accessible for user to make use of it outside the tutorial, as they should rather get the latest yeast-GEM version
- 


src/geckomat/model_adapter/ModelAdapter.m
```diff
@@ -18,7 +18,7 @@
         function model = manualModifications(obj,model) %default is to do nothing
```
ae-tafur, line 18: This line is not need any more, can be removed since is an artifact of the v2. Initial try to move/refactor `manualModifications.m`

ae-tafur (approved):

PR #335, opened 2023-07-05
docs: futher polishing

mihai-sysbio:

This PR is used to highlight a few places where the tutorials can be further improved/simplified. Please feel free to push commits to this branch.

### DIFF
--- a/tutorials/full_ecModel/protocol.m
+++ b/tutorials/full_ecModel/protocol.m
@@ -305,8 +305,8 @@ saveEcModel(ecModel,'ecYeastGEM_stage3.yml');
 saveEcModel(ecModel,'ecYeastGEM.yml');
 
 %% STAGE 4 integration of proteomics data into the ecModel.
-%ecModel=loadEcModel('ecYeastGEM_stage3.yml'); % Uncomment if you want to
-%reload model.
+% Uncomment the line below if you want to reload the model.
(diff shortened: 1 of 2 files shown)


mihai-sysbio:
It takes less than 10 min to run the light tutorial now, great job 🌟 (I know it's using a lot of pre-computed values in the TSV files, but we do provide ways to reproduce them directly in the tutorial)

mihai-sysbio:
The order of installing GECKO and checking the RAVEN installation is different in the 2 tutorials. Any insight into which is preferred?

mihai-sysbio:
In STEP 22 of the full tutorial
> sol = solveLP(ecModel)

prints a growth rate of 0.0887, which very close to 0.0889 but still different. Any reactions?

mihai-sysbio:
I've ran both tutorials several times in less than 1 hour in total, very convenient 🤩

edkerk:
> The order of installing GECKO and checking the RAVEN installation is different in the 2 tutorials. Any insight into which is preferred?

Best to have RAVEN first, as GECKO will do a check that RAVEN can be found, I'll swap it.

> In STEP 22 of the full tutorial
> 
> > sol = solveLP(ecModel)
> 
> prints a growth rate of 0.0887, which very close to 0.0889 but still different. Any reactions?

I'll slightly rephrase it to avoid mentioning precise numbers.

edkerk (approved):

Issue #334, opened 2023-07-04
readDLKcatOutput reports non-matching metabolites


HossFir:
I generated the new DLKcat.tsv using `writeDLKcatInput(ecModel,[],[],[],[],true);` and subsequently, `runDLKcat();`

It generated the new DLKcat.tsv with my organism information. However, I saw an error on the following line:

https://github.com/SysBioChalmers/GECKO/blob/63785559f6c7ebdb908386da92d56a034571c449/tutorials/full_ecModel/protocol.m#L145

It stated that _Not all substrates from DLKcat output can be found in model.metNames. DLKcat was likely run with an input file that was generated from another ecModel._

I had no such error on the previous version 3.0.2.

edkerk:
Version 3.1.0 has not seen any change in parsing metabolite names, and if anything should result in _less_ metabolites by more thorough filtering of currency metabolites. Are you sure you have the right model file loaded, and direct to the right DLKcat.tsv file? I'll work on `readDLKcatOutput` giving a more complete error message, reporting the non-matching substrates,

For the moment you can get more information by running:

```
dbstop if error
kcatList_DLKcat = readDLKcatOutput(ecModel); 
```
and when the error is reported, run the following lines:
```
tmp = ismember(subs,model.metNames));
nonMatchingMets = subs(tmp);
```
Then, nonMatchingMets has the names of metabolites from DLKcat.tsv that cannot be found in `model.metNames`.

Issue #332, opened 2023-07-03
Flux comparison using GLPK & CPLEX


Yu-sysbio:
I used GLPK and CPLEX (via COBRA) to run Step 29 but got quite different plots (see attached [plots.zip](https://github.com/SysBioChalmers/GECKO/files/11933818/plots.zip)). The deviation comes from the comparison between extremely low fluxes. While the plots look different, the general conclusion still holds: the simulated objective fluxes are the same, and there is no big difference between mapped fluxes simulated by light and full ecModels.

I have an issue with the license of GUROBI currently and thus do not have the corresponding result for comparison. But I suppose that the plot in the manuscript was generated using GUROBI, which looks much better than the other two solvers. Maybe we could mention in the manuscript that all the anticipated results are obtained with GUROBI (I also got a slightly different FVA plot using GLPK). And maybe only plot high absolute fluxes (e.g. > 1e-6) as the extremely low fluxes make no sense.

Below I also put what I got in the Command Window using these two solvers:

**GLPK**:

> Comparison of duration light vs. full ecModel
> ecModel reconstruction: 91% (11 vs 12 seconds)
> WARNING: Could not solve the problem of minimizing the sum of fluxes. Uses output from original problem
> WARNING: Could not solve the problem of minimizing the sum of fluxes. Uses output from original problem
> FBA: 90% (1.34 vs 1.50 seconds)
> Mapping fluxes: 60% (0.037 vs 0.063 seconds)
> Growth rate that is reached: 0.0252 vs 0.0252

**CPLEX**:

> Comparison of duration light vs. full ecModel
> ecModel reconstruction: 78% (11 vs 14 seconds)
> FBA: 68% (0.61 vs 0.90 seconds)
> Mapping fluxes: 39% (0.038 vs 0.099 seconds)
> Growth rate that is reached: 0.0252 vs 0.0252

**I hereby confirm that I have:**
- [x] Done this analysis in the `feat/compareLightVsFull` branch of the repository
- [x] Checked that a similar issue does not exist already

edkerk:
It might be good to double-check which reactions have those very low fluxes.

But it indeed seems to be an artifact of the solver and not the true difference of the flux distributions. I can modify `plotlightVSfull` to set these fluxes that are below the solver cut-off to 0.

In the Materials section is mentioned that Gurobi is strongly recommended.

edkerk:
Please review #331 to see if the plots look more similar.

[...]

PR #336, opened 2023-07-09
feat: flexEnz instead of flexProt + more detailed errors

edkerk:
- Features: 
  - `readDLKcatOutput` error gives list of non-matching metabolites and reactions (solves #334)
  - `mapRxnsToConv` error if empty flux vector is used as input (solves #332)
- Refactor:
  - Use `flexibilizeEnzConcs` as function name for flexibilizing enzyme concentrations in the model.


ae-tafur (approved):
Looks good

PR #337, opened 2023-07-10
feat: Add new rxns to ecModel & fix: ecFSEOF

ae-tafur:

- Add new reactions to an ecModel (`addNewRxnsToEC`): This function is useful to simulate metabolic manipulations done to an organism such as integration of new genes that add a new reaction/pathway. After add the new reactions, `setKcatForReactions` should be run to add the kcat to the reactions, and subsequently `applyKcatConstraints`.
- Fix consistency names in ecFSEOF functions
- Feat add progress bar to `ecFSEOF`
- Integrate `ecFluxScanning` into `ecFSEOF`


edkerk:
Looks good, will test it out. A few things:
- Probably good to mention about `setKcatForReactions` and `applyKcatConstraints` in the function description.
- The old function name (`getEngineeredModel`) is occassionally mentioned.
- Should `newEnzymes` not be an optional field? The gene/enzyme might already be present. Together with that, there should be a check whether the gene/enzyme not already exist (ok, `addGenesRaven` does that check, but its warning message will seem strange as the user will not have defined a genesToAdd structure themselves, nor does it handle enzymes).

edkerk:
Sorry, I meant to have `newEnzymes` is an optional _input_ for running `addNewRxnsToEC.m`. If I know that an enzyme is present, do I really have to provide `newEnzymes` information.

I now see that you indeed have a check that errors if one of the enzymes is present. Another way to deal with that is to check if the enzyme is present, if present remove it from the list of `newEnzymes`, and then only run steps 2, 4, 5 and partial step 6 for any truly new enzymes. But if I now indeed add a reaction with only existing enzymes, I will get an error that these should be removed from `newEnzymes`, after which I will have to provide an empty `newEnzymes`, which will then error at these steps when pseudometabolite is added etc.

ae-tafur:
> Sorry, I meant to have `newEnzymes` is an optional _input_ for running `addNewRxnsToEC.m`. If I know that an enzyme is present, do I really have to provide `newEnzymes` information.
> 
> I now see that you indeed have a check that errors if one of the enzymes is present. Another way to deal with that is to check if the enzyme is present, if present remove it from the list of `newEnzymes`, and then only run steps 2, 4, 5 and partial step 6 for any truly new enzymes. But if I now indeed add a reaction with only existing enzymes, I will get an error that these should be removed from `newEnzymes`, after which I will have to provide an empty `newEnzymes`, which will then error at these steps when pseudometabolite is added etc.

OK, Initially, my idea was to add reactions that are not part of the model because new genes are incorporated to the organism. So, I was assuming that if a gene is already part of the genome of that model, the reaction should be there. But I think I got you point.

edkerk (approved):

PR #395, opened 2024-10-19
chore/feat: dlkcat docker multiarch

mihai-sysbio:
This PR adds an ARM-compatible Docker image of `dlkcat`.

- [x] This PR has `develop` as target branch, and will be resolved with a *squash-merge*.
- [ ] This PR has `main` as target branch, and will be resolved with a *merge commit*.


mihai-sysbio:
Once all the images are tested, it would be good to change the call `dlkcat-gecko` to `0.1-multiarch` rather than just `0.1`, but the old `0.1` image should not be deleted.

PR #418, opened 2026-06-17
Fix a bug where DLKcat cannot write its own results

emilpaulitz:
The bug surfaces rarely but does occur and the fix should not impact the behavior negatively.

- Fixes:
  - `src/geckomat/gather_kcats/runDLKcat.m` sometimes failed to write the DLKcat output because the option to overwrite the input file is not specified.

- [ ] This PR has `develop` as target branch, and will be resolved with a *squash-merge*.
- [x] This PR has `main` as target branch, and will be resolved with a *merge commit*.


edkerk:
Nice fix! Could you please apply it in the `develop` branch and target the `develop` branch, instead of `main` branches?

edkerk:
Thanks for the fix @emilpaulitz! The change has been applied to the `develop4` branch in #421, which is the correct target for bug fixes. Closing this PR.

PR #422, opened 2026-08-18
Fix: KEGG parsing whitespace bug and infinite retry on download failure

ae-tafur:

While working with *Escherichia coli*, the downloaded `kegg.tsv` file had errors and did not keep the expected structure due to a stray leading newline in the KEGG entries after parsing. This PR fixes that issue, along with a couple of related improvements found while investigating it.

- Fixes:
  - `downloadKEGG` left a stray leading newline in `keggData` entries after
    splitting on `///`, which broke the `startsWith(...,'ENTRY ')` checks used
    to filter out genes without AASEQ or UniProt fields, letting invalid
    entries slip into the final KEGG database table
  - `downloadKEGG` retry loop for `webread` had no attempt limit, causing the
    function to hang indefinitely if KEGG was down or the URL was invalid,
    with the original error silently discarded
- Performance:
  - `downloadKEGG` used `\s*` instead of `\s+` when stripping whitespace from
    protein sequences, causing unnecessary zero-length matches at every
    position in the string

- [X] This PR has `develop` as target branch, and will be resolved with a *squash-merge*.
- [ ] This PR has `main` as target branch, and will be resolved with a *merge commit*.

### DIFF
--- a/src/geckomat/get_enzyme_data/loadDatabases.m
+++ b/src/geckomat/get_enzyme_data/loadDatabases.m
@@ -146,13 +146,22 @@ for i = 1:queries
     end
     url      = ['http://rest.kegg.jp/get/' keggID ':' strjoin([gene_id{firstIdx:lastIdx}],['+' keggID ':'])];
 
-    retry = true;
-    while retry
+    maxRetries = 5;
+    attempt    = 0;
+    success    = false;
+    while ~success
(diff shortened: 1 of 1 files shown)


edkerk (commented):
Works as intended!

edkerk (approved):

Issue #449, opened 2026-08-28
readDLKcatOutput.m matches DLKcat substrate names against model.metNames case-sensitively


edkerk:
## Bug

\`readDLKcatOutput.m\` checks that every DLKcat-output substrate name appears in the model:

\`\`\`matlab
matchMets = ismember(subs,model.metNames);
\`\`\`

\`ismember\` on cell arrays of char defaults to case-sensitive comparison. An SBML loader that normalises metabolite-name capitalisation differently than whatever produced the DLKcat input causes spurious "substrate not found" failures on names that are the same metabolite, just differently cased.

## Expected fix

\`\`\`matlab
matchMets = ismember(lower(subs),lower(model.metNames));
\`\`\`

## Parity

geckopy's \`read_dlkcat_output\` already does the case-insensitive comparison, and its own docstring documents the MATLAB gap directly:

> MATLAB-COMPAT: GECKO MATLAB's substrate-name match against \`model.metNames\` is case-sensitive (\`ismember\` default). geckopy uses case-insensitive matching, which is more lenient and avoids false-positive failures when SBML loaders differ in capitalization.

## Notes

Diagnosed and already fixed once on the long-abandoned \`feat/geckopy-compat\` branch (commit \`0b6b47f\`, never merged — that branch is 99 commits behind \`develop4\`). Will follow up with a PR against current \`develop4\`.

edkerk:
Verified fixed on `develop4`: `readDLKcatOutput.m` now matches case-insensitively (`ismember(lower(subs),lower(model.metNames))`). Also recorded on the raven-gecko-parity ledger as closed by GECKO#452.

Issue #451, opened 2026-08-28
makeEcModel.m does not sort ec.genes, so ecModel.ec struct order is not reproducible across runs


edkerk:
## Gap

\`makeEcModel.m\`'s \`ec.genes\` (and, in lockstep, \`ec.enzymes\`/\`ec.mw\`/\`ec.sequence\`) end up in whatever order the underlying data-gathering step happens to produce them, rather than a stable, sorted order. Two runs against the same input can therefore produce ecModels whose \`ec\` struct fields are permuted relative to each other even though the model is otherwise identical — makes diffing/comparing ecModels across runs or machines harder than it needs to be.

## Expected fix

Sort \`ec.genes\` alphabetically once populated, permuting \`ec.enzymes\`/\`ec.mw\`/\`ec.sequence\` (and anything else indexed the same way) identically, so the \`ec\` struct has a stable, comparable order.

## Parity

Not a MATLAB-vs-Python behavioural divergence — geckopy has no explicit gene-ordering guarantee in \`make_ec_model\` either (nothing in its \`MATLAB-COMPAT\` notes addresses this), so there's no specific Python behaviour to match. This is a MATLAB-side determinism/reproducibility improvement on its own terms, not a parity fix.

## Notes

Implemented once already on the long-abandoned \`feat/geckopy-compat\` branch (commit \`0b6b47f\`, never merged). Will follow up with a PR against current \`develop4\`, bundled with a couple of other small fixes found the same way.

edkerk:
Correction: this is a confirmed parity fix, not just a MATLAB-side improvement as I originally wrote above.

I checked \`make_ec_model.py\`'s top level for gene-sorting logic and found none, and concluded geckopy had no equivalent behaviour to match. I was checking the wrong file — the actual stage-7 logic lives in \`pipeline/populate_ec.py\`, which already sorts and says so directly:

> MATLAB-COMPAT: GECKO MATLAB iterates model.genes in their original model order. geckopy iterates them in alphabetical order so that ec.genes / ec.enzymes / ec.mw / ec.sequence are deterministic regardless of how the SBML was loaded.

So this closes an already-documented, already-known divergence rather than adding a MATLAB-only nicety. (raven-gecko-parity's own ledger already had this catalogued too — I should have checked there first.)

edkerk:
[...]

Issue #450, opened 2026-08-28
fuzzyKcatMatching.m carries a dead forceWClvl==1 branch


edkerk:
## Cleanup, not a behavioural bug

\`fuzzyKcatMatching.m\` has:

\`\`\`matlab
%Apply force wildcard level
while forceWClvl > 0
    eccodes=regexprep(eccodes,'(.)*(\.\d+)(\.-)*$','$1\.-$3');
    forceWClvl = forceWClvl - 1;
end
if forceWClvl == 1
    eccodes = regexprep(eccodes,'.*','-\.-\.-\.-');
end
\`\`\`

The preceding \`while\` loop unconditionally decrements \`forceWClvl\` to exactly \`0\` before falling through (it only exits once \`forceWClvl > 0\` is false). The following \`if forceWClvl == 1\` can therefore never be true — dead code, no observable effect on output either way.

## Expected fix

Delete the dead \`if forceWClvl == 1 ... end\` block.

## Notes

No parity implication (unreachable code affects nothing on either side). Spotted and removed once already on the long-abandoned \`feat/geckopy-compat\` branch (commit \`0b6b47f\`, never merged). Will follow up with a PR against current \`develop4\`, bundled with a couple of other small fixes found the same way.

edkerk:
Verified fixed on `develop4`: `fuzzyKcatMatching.m`'s force-wildcard block is now just the `while forceWClvl > 0` loop -- the dead `if forceWClvl == 1` branch after it is gone.

Issue #448, opened 2026-08-28
saveEcModel.m's default filename has no .yml extension


edkerk:
## Bug

\`saveEcModel.m\` defaults to:

\`\`\`matlab
if isempty(filename)
    filename = 'ecModel';
end
\`\`\`

With no extension, this falls through to \`writeYAMLmodel\`, which writes a file with no \`.yml\` suffix — every default-named save produces an extensionless file.

## Expected fix

\`\`\`matlab
if isempty(filename)
    filename = 'ecModel.yml';
end
\`\`\`

## Parity

geckopy's \`save_ec_model\` already gets this right — its default is \`"ecModel.yml"\`, and it additionally validates that any *explicit* filename ends in \`.yml\`/\`.yaml\`, raising otherwise. MATLAB doesn't need to go that far to fix this bug, but the target default value is confirmed against the Python side.

## Notes

Diagnosed and already fixed once on the long-abandoned \`feat/geckopy-compat\` branch (commit \`0b6b47f\`, never merged — that branch is 99 commits behind \`develop4\`). Will follow up with a PR against current \`develop4\`.

edkerk:
Not actually a live bug against current develop4 — re-checked while writing the fix. RAVEN's
writeYAMLmodel.m already guards this:

    if ~endsWith(fileName,{'.yml','.yaml'})
        fileName = strcat(fileName,'.yml');
    end

saveEcModel.m passes its (extensionless) default filename straight through to writeYAMLmodel,
which appends `.yml` unconditionally regardless of what saveEcModel itself defaults to. The
default-filename round trip (save then load) already produces and reads back
`models/ecModel.yml` correctly today; there's no user-visible bug to fix.

The original diagnosis on feat/geckopy-compat was accurate for whatever RAVEN version that
branch was built against 6+ months ago, but doesn't hold against RAVEN's current
writeYAMLmodel.m. Closing without a code change.

PR #452, opened 2026-08-28
fix: case-insensitive DLKcat substrate match, stable ec.genes order, dead code

edkerk:
## Summary

Three independent small fixes, each with its own tracking issue:

- **readDLKcatOutput.m** (#449): substrate-name matching against `model.metNames` was case-sensitive, causing spurious "substrate not found" failures when an SBML loader normalises capitalisation differently than whatever produced the DLKcat input. Now case-insensitive, matching geckopy's already-lenient `read_dlkcat_output`, whose own docstring documents this exact MATLAB gap.
- **makeEcModel.m** (#451): `ec.genes` (and `ec.enzymes`/`ec.mw`/`ec.sequence`, permuted in lockstep) now sort alphabetically, so the `ec` struct has a stable, comparable order across runs instead of whatever order the underlying data-gathering step happened to produce. Not a parity fix — geckopy's `make_ec_model` has no equivalent gene-ordering guarantee to match either, so this is a MATLAB-side determinism improvement on its own terms.
- **fuzzyKcatMatching.m** (#450): removed a dead `if forceWClvl == 1` block — the preceding `while` loop always decrements `forceWClvl` to exactly 0 first, so the branch could never fire. No behavioural change on either side.

A fourth reported bug (`saveEcModel.m`'s default filename lacking a `.yml` extension) turned out not to reproduce against current `develop4`: RAVEN's `writeYAMLmodel.m` already appends `.yml` unconditionally whenever it's missing, so the save/load round trip already produces and reads back `ecModel.yml` correctly today. Closed without a code change (#448) rather than ship a fix for a non-bug.

All four were originally diagnosed on the long-abandoned `feat/geckopy-compat` branch (commit `0b6b47f`, never merged — 99 commits behind `develop4`) and re-verified against current `develop4` here rather than merging that branch.

## A separate, more serious finding along the way

While writing the regression test for the `readDLKcatOutput.m` fix, reverting the fix to confirm the test failed correctly surfaced an **unrelated, currently-live crash**: `dispEM`, a RAVEN utility function, was removed from RAVEN in [SysBioChalmers/RAVEN#659](https://github.com/SysBioChalmers/RAVEN/pull/659) (merged 2026-06-18, already on `develop3`) and replaced with native `warning`/`error` + a new `ravenList` formatter. GECKO's `develop4` still calls the now-nonexistent `dispEM` in four places:

- `src/geckomat/gather_kcats/readDLKcatOutput.m` (2 call sites)
- `src/geckomat/get_enzyme_data/loadDatabases.m` (1)
- `src/geckomat/utilities/addNewRxnsToEC.m` (1)
- `src/geckomat/utilities/getSubsetEcModel.m` (1)

Each of these crashes with `Undefined function 'dispEM'` instead of reporting its intended warning/error whenever it reaches that branch — i.e. exactly when something is already wrong and the function is trying to tell the caller. Not fixed in this PR (out of scope for what was asked here); will follow up separately.

## Test plan

- [x] Added `testReadDLKcatOutputSubstrateMatchIsCaseInsensitive_tc0035` and `testMakeEcModelSortsEcGenes_tc0036` to `geckoCoreFunctionTests.m`.
- [x] Verified each new test fails against the pre-fix code and passes against the fix.
- [x] Full local suite: 36/36 passed.

Closes #449, #450, #451.

### DIFF
--- a/src/geckomat/change_model/makeEcModel.m
+++ b/src/geckomat/change_model/makeEcModel.m
@@ -282,10 +282,13 @@ elseif ~isempty(noUniprot)
         'data (e.g. from a different proteome, make sure you first delete '...
         'the existing data/uniprot.tsv file.\n']);
 end
-ec.genes        = model.genes(Lia); %Will often be duplicate of model.genes, but is done here to prevent issues when it is not.
-ec.enzymes      = uniprotDB.ID(Locb(Lia));
-ec.mw           = uniprotDB.MW(Locb(Lia));
-ec.sequence     = uniprotDB.seq(Locb(Lia));
+%Sort alphabetically so ec.genes has a stable order across runs.
+[ec.genes,sIdx] = sort(model.genes(Lia)); %Will often be duplicate of model.genes, but is done here to prevent issues when it is not.
+LocbSorted      = Locb(Lia);
+LocbSorted      = LocbSorted(sIdx);
+ec.enzymes      = uniprotDB.ID(LocbSorted);
+ec.mw           = uniprotDB.MW(LocbSorted);
+ec.sequence     = uniprotDB.seq(LocbSorted);
 %Additional info
 ec.concs        = nan(numel(ec.genes),1); % To be filled with proteomics data when available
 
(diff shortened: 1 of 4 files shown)


edkerk:
Correction to the makeEcModel.m bullet above: this closes a confirmed, already-documented MATLAB-vs-Python divergence, not a MATLAB-only improvement with nothing to match. See the follow-up on #451 — geckopy's \`pipeline/populate_ec.py\` already sorts \`ec.genes\` alphabetically and says so directly in a \`MATLAB-COMPAT\` note; I'd checked the wrong file (\`make_ec_model.py\`'s top level) when I said otherwise.

PR #471, opened 2026-09-01
docs: remove cross-language and code-history commentary

edkerk:
## Summary
Comments and docstrings described current behavior in comparison to geckopy, and narrated how the code changed over time (issue/PR references, "used to" framing, before/after comparisons). Rewritten to describe only the current MATLAB behavior and its rationale.

- `getReactionsFromEnzyme.m`: dropped the geckopy comparison and issue reference from the Notes docstring, kept the current-behavior facts (empty-on-unmatch, case-insensitive matching).
- `geckoCoreFunctionTests.m`: rewrote test rationale comments (tc0014, tc0016, tc0018, tc0047-tc0054) to describe only what each test currently verifies, with issue citations and "used to be broken" narratives removed. Also renamed `testCalculateMWReconciledWithGeckopy_tc0046` -> `testCalculateMWResidueMassHandling_tc0046` (only the geckopy reference was in the name; the body was already clean). Test logic/assertions untouched throughout.
- `fuzzyKcatMatching.m`: minor rewording, no functional change.
- `loadBRENDAdata.m`: minor rewording (no functional change), **plus a real fix**: the inline comment on the derived SA*MW kcat value said `%[1/hr]`; it's actually `%[1/s]` (SA is scaled to mmol/(s*g), MW to g/mmol, so SA*MW = 1/s). Confirmed against geckopy's `brenda_loader.py`, which computes the identical product and documents it as 1/s. The value itself was always correct -- only the comment was wrong.

## Test plan
- [x] Full `geckoCoreFunctionTests.m` suite, run solo (no concurrent MATLAB processes) — 54 passed, 0 failed. (protocol.m is a tutorial script, not exercised by the test suite.)

### DIFF
--- a/src/geckomat/change_model/getReactionsFromEnzyme.m
+++ b/src/geckomat/change_model/getReactionsFromEnzyme.m
@@ -24,9 +24,8 @@ function [rxns, kcat, idx, rxnNames, grRules] = getReactionsFromEnzyme(ecModel,p
 % Notes
 % -----
 % An unmatched proteinId is not an error: all five outputs are returned
-% empty, with no warning. geckopy's get_reactions_from_enzyme matches
-% this function's case-insensitive matching, but deliberately raises
-% instead of returning empty on an unmatched id.
+% empty, with no warning. Matching against ecModel.ec.enzymes is
+% case-insensitive.
 %
 % Examples
 % --------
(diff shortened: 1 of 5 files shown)


ae-tafur:
@edkerk, a suggestion for `protocol.m` — try using a Live Script instead. Newer MATLAB versions let you save them as `.m` files, so you still get clean version control diffs without losing the Live Script formatting. It'll give users a much clearer view of the protocol.

